PR #1, opened 2020-06-26
HelmRelease v2alpha1 spec

hiddeco:


docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+		// +optional
+		Enable bool `json:"enable,omitempty"`
+
+		// OnConditionReason defines on what condition reasons a Helm test
+		// should be performed. Defaults to ['InstallSucceeded', 'UpgradeSucceeded'].
+		// +optional
+		OnConditionReason []string `json:"onConditionReason,omitempty"`
+
+		// Timeout for the Helm test action. Defaults to the main
+		// 'Timeout' duration.
```
seaneagan, line 113: I think an explicit default, matching that of the `helm test` command (300s) which is what is currently implemented in the Helm Operator is more intuitive. I don't think the timeout for the release itself it closely related to the timeout for its tests.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+	TargetNamespace string `json:"targetNamespace,omitempty"`
+
+	// SourceRef of the Helm chart source.
+	// +required
+	SourceRef corev1.TypedLocalObjectReference `json:"sourceRef"`
+
+	// DependsOn holds a list of HelmReleases that must be ready before this
+	// release can be installed.
+	// +optional
+	DependsOn []string `json:"dependsOn,omitempty"`
```
seaneagan, line 31: Noticed we're using `string` here to refer to `HelmRelease`s, while using `corev1.TypedLocalObjectReference` above to reference `HelmChart`s.  Why this difference?

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+
+		// DisableHooks prevents hooks from running during rollback. Defaults
+		// to false.
+		// +optional
+		DisableHooks bool `json:"disableHooks,omitempty"`
+
+		// Force resource update through delete/recreation if needed. Defaults
+		// to false.
+		// +optional
+		Force bool `json:"force,omitempty"`
```
seaneagan, line 155: Does this configure the upgrade or the rollback or both? Presumably ideal would be to have this be for rollback and to add an additional `Force` field under `Upgrade` for upgrades.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+
+		// Recreate performs pod restarts for the resource if applicable. Defaults
+		// to false.
+		// +optional
+		Recreate bool `json:"recreate,omitempty"`
```
seaneagan, line 160: This option is removed in Helm 3.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+		// Timeout for the Helm test action. Defaults to the main
+		// 'Timeout' duration.
+		// +optional
+		Timeout *metav1.Duration `json:"timeout,omitempty"`
+	} `json:"test,omitempty"`
+
+	// Rollback holds the configuration for enabling and performing Helm
+	// rollback actions.
+	// +optional
+	Rollback *struct {
```
seaneagan, line 121: A similar `Uninstall` key would be useful as well, see https://github.com/fluxcd/helm-operator/issues/388. In addition to what's mentioned there, an `Uninstall.OnConditionReason` would be needed in order to ignore test failures after installs.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+
+A **helmrelease** object defines the source of a Helm chart by referencing an object
+managed by [source-controller](https://github.com/fluxcd/source-controller), the interval
+reconciliation should happen at, and a set of options to control the settings of the
+automated Helm actions that are being performed.
+
+```go
+type HelmReleaseSpec struct {
+	// ReleaseName used for the Helm release. Defaults to a composition of
+	// '[TargetNamespace-]Name'.
```
seaneagan, line 15: Why add the `TargetNamespace`? If it's to ensure prevent HelmReleases accidentally using the same release name when setting `TargetNamespace`, then wouldn't it make more sense to just always default to `metadata.namespace-metadata.name` of the HelmRelease since that is always unique to a HelmRelease? Whereas two HelmReleases could choose the same `TargetNamespace`. I wonder if this is even needed though, especially if we're keeping something like the [antecedent annotation](https://docs.fluxcd.io/projects/helm-operator/en/stable/helmrelease-guide/reconciliation-and-upgrades/#the-antecedent-annotation) implemented.

docs/spec/v2alpha1/helmreleases.md
hiddeco: This is planned, see developments in #2. This POC also no longer relies on condition reasons but on condition types and statuses.

docs/spec/v2alpha1/helmreleases.md
hiddeco: Because a `TypedLocalObjectReference` can refer to other types than `HelmRelease` resources, and you should only be able to depend on things within the same namespace, which makes this a simple `metadata.Name` list.

docs/spec/v2alpha1/helmreleases.md
hiddeco: It is to prevent `HelmRelease` resources from colliding, see the explanation [here](https://docs.fluxcd.io/projects/helm-operator/en/stable/helmrelease-guide/release-configuration/#specifying-a-release-name).

> I wonder if this is even needed though, especially if we're keeping something like the antecedent annotation implemented.

We no longer annotate resources, nor do we try to be smart with Helm state (e.g. dry-runs). We simply record the revision of the latest Helm release as deployed by the controller in the status object of the resource and use this to determine if we need to perform an upgrade.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,276 @@
...
+
+```go
+const (
+	// ReadyCondition represents the fact that the HelmRelease
+	// has been successfully enrolled on the cluster.
+	ReadyCondition string = "Ready"
+)
+```
+
+Status condition reasons:
```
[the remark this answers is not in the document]
stefanprodan: Yes like source and kustomize controllers, the helm controller will emit events for the notification controller to dispatch: https://github.com/fluxcd/notification-controller/blob/master/docs/spec/v1alpha1/event.md

docs/spec/README.md
```diff
@@ -0,0 +1,66 @@
...
+could support in the future.
+
+Reconciliation of the `HelmRelease` runs on-a-schedule and can be triggered manually by a
+cluster admin, or automatically by a source event such as a Helm chart revision change.
+
+When a custom resource is deleted from the cluster, the controller's garbage collector removes
+the associated `HelmChart` and uninstalls the Helm release. Deleting a suspended resource does not
+trigger a Helm uninstall.
+
+Alerting can be configured with a Kubernetes custom resource that specifies a webhook address, and a
```
stefanprodan, line 50: We should add a link to notification-controller.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -0,0 +1,353 @@
...
+	// reconciliation outside of the specified schedule.
+	ReconcileAtAnnotation string = "helm.fluxcd.io/reconcileAt"
+)
+```
+
+On-demand execution example:
+
+```bash
+kubectl annotate --overwrite helmrelease/podinfo helm.fluxcd.io/reconcileAt="$(date +%s)"
+```
```
stefanprodan, line 353: Example section is missing.

Issue #6, opened 2020-07-09
Emit events


stefanprodan:
Helm controller should emit events whenever a release status changes and integrate with [notification controller](https://github.com/fluxcd/notification-controller).

PR #7, opened 2020-07-09
Implement event recording

stefanprodan:
Changes:
- emit Kubernetes events when a release status changes
- forward events to notification controller
- use Alpine as container image
- add myself to maintainers 
- update overview diagram

Fix: #6


main.go
```diff
@@ -66,6 +70,16 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(!logJSON)))
 
+	var eventRecorder *recorder.EventRecorder
+	if eventsAddr != "" {
+		if er, err := recorder.NewEventRecorder(eventsAddr, "kustomize-controller"); err != nil {
```
hiddeco, line 75:
```suggestion
		if er, err := recorder.NewEventRecorder(eventsAddr, "helm-controller"); err != nil {
```

PR #23, opened 2020-07-16
Refactor garbage collection to finalizer

hiddeco:
Ref: https://github.com/fluxcd/toolkit/issues/89

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -80,6 +80,38 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 
 	log := r.Log.WithValues("controller", strings.ToLower(v2.HelmReleaseKind), "request", req.NamespacedName)
 
+	// Examine if the object is under deletion
+	if hr.ObjectMeta.DeletionTimestamp.IsZero() {
+		// The object is not being deleted, so if it does not have our finalizer,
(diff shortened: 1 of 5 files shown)


controllers/helmrelease_controller.go
```diff
@@ -80,6 +80,38 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+		}
+	} else {
+		// The object is being deleted
+		if containsString(hr.ObjectMeta.Finalizers, v2.HelmReleaseFinalizer) {
+			// Our finalizer is still present, so lets handle garbage collection
+			if err := r.gc(ctx, log, hr); err != nil {
+				r.event(hr, hr.Status.LastAttemptedRevision, recorder.EventSeverityError, fmt.Sprintf("garbage collection for deleted resource failed: %s", err.Error()))
+				// Return the error so we retry the failed garbage collection
+				return ctrl.Result{}, err
+			}
```
stefanprodan, line 103: Can we send an event like "Helm uninstall succeeded"?

hiddeco: There are two options: send an info event here with the fact that the garbage collection for the deleted resource succeeded.

Or broadcast the messages from within the `gc` method for the individual garbage collection actions.

stefanprodan: I would opt for a single event at the end.

PR #27, opened 2020-07-17
Uninstall fixes

seaneagan:
* Ensure Status.LastReleaseRevision is 0 after uninstall. This
  could be nil instead, but since 0 is never a valid release revision,
  it seems sufficient.
* Ensure uninstall leads to ready condition being false. This has
  no effect now since uninstall is only triggered on install failure,
  but will once other triggers (such as test failure) are supported.


hiddeco:
Thanks for your contributions :tulip:

My weekend is about to start, so I have self-assigned them for review on Monday.

PR #31, opened 2020-07-21
Fix: use fetched HelmChart object for update

### LINKED ISSUE #30: Chart update fails
When changing the chart version, HC can't update the HelmChart:

```
helmcharts.source.fluxcd.io "gitops-system-podinfo" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

hiddeco:
Fixes #30

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -243,16 +243,16 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, false, err
 	}
+	hc := helmChartFromTemplate(*hr)
 	switch {
 	case apierrors.IsNotFound(err):
-		hc := helmChartFromTemplate(*hr)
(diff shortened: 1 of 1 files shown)


controllers/helmrelease_controller.go
```diff
@@ -243,16 +243,19 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
...
 	case apierrors.IsNotFound(err):
-		hc := helmChartFromTemplate(*hr)
 		if err = r.Client.Create(ctx, hc); err != nil {
 			return nil, false, err
 		}
 		hr.Status.HelmChart = chartName.String()
 		return nil, false, nil
 	case helmChartRequiresUpdate(*hr, helmChart):
-		hc := helmChartFromTemplate(*hr)
+		helmChart.Spec.Name = hc.Spec.Name
```
stefanprodan, line 255: I would do `helmChart.Spec = hc.Spec` so we don't need to update this code if we add more fields later on.

PR #54, opened 2020-08-19
Allow overwrites of map values with flat values

hiddeco:
After taking a closer look at the Helm code, this is the behaviour
we are actually after, given we want to mimic the merging behaviour
of the -f flag available to Helm install and/or upgrade commands.


stefanprodan (approved):
LGTM

Issue #42, opened 2020-07-23
Providing values from ConfigMap and Secret resources


hiddeco:
To be able to provide values by other means than in-lining them in the `HelmRelease` resource. We want to provide users with the option to reference to values in `ConfigMap` and `Secret` resources.

## Spec changes

```diff
diff --git a/docs/spec/v2alpha1/helmreleases.md b/docs/spec/v2alpha1/helmreleases.md
index 5c13e45..086f624 100644
--- a/docs/spec/v2alpha1/helmreleases.md
+++ b/docs/spec/v2alpha1/helmreleases.md
@@ -70,6 +70,10 @@ type HelmReleaseSpec struct {
 	// +optional
 	Uninstall Uninstall `json:"uninstall,omitempty"`
 
+	// ValuesFrom holds references to values in Secrets and ConfigMaps.
+	// +optional
+	ValuesFrom []ValuesReference `json:"valuesFrom,omitempty"`
+
 	// Values holds the values for this Helm release.
 	// +optional
 	Values apiextensionsv1.JSON `json:"values,omitempty"`
@@ -96,6 +100,30 @@ type HelmChartTemplate struct {
 	Interval *metav1.Duration `json:"interval,omitempty"`
 }
 
+// ValuesReference contains enough information to let you locate the
+// values reference at namespace level, and where the (sub)set of
+// values should be merged at.
+type ValuesReference struct {
+	// Kind of the values referent.
+	// +kubebuilder:validation:Enum=Secret,ConfigMap
+	// +required
+	Kind string `json:"kind"`
+
+	// Name of the values referent.
+	// +required
+	Name string `json:"name"`
+
+	// ValuesKey is the key in the referent the values can be found at.
+	// Defaults to 'values.yaml'.
+	// +optional
+	ValuesKey string `json:"valuesKey,omitempty"`
+
+	// TargetPath is the YAML dot notation path the values should be merged at.
+	// Defaults to 'None', which results in the values getting merged at the root.
+	// +optional	
+    TargetPath string `json:"targetPath,omitempty"`
+}
+
 // Install holds the configuration for Helm install actions.
 type Install struct {
 	// Timeout is the time to wait for any individual Kubernetes operation (like Jobs
```

## Changes vs Helm Operator's `valuesFrom`

* Values from external sources, for example an URL, are no longer supported.

[...]

PR #55, opened 2020-08-19
Support targeting a path in ValuesReferences

hiddeco:
Ref: #42

### DIFF
--- a/api/v2alpha1/reference_types.go
+++ b/api/v2alpha1/reference_types.go
@@ -50,10 +50,16 @@ type ValuesReference struct {
 	// +required
 	Name string `json:"name"`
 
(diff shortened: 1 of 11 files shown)


api/v2alpha1/reference_types.go
```diff
@@ -50,10 +50,16 @@ type ValuesReference struct {
 	// +required
 	Name string `json:"name"`
 
-	// ValuesKey is the key in the referent the values can be found at.
+	// ValuesKey is the key in the referent the values or value can be found at.
```
stefanprodan, line 53:
```suggestion
	// ValuesKey is the data key where the values.yaml or a specific value can be found at.
```

stefanprodan (approved):
LGTM

PR #62, opened 2020-08-31
Support charts from GitRepository sources

hiddeco:
Ref: https://github.com/fluxcd/source-controller/issues/56


stefanprodan (approved):
LGTM

PR #67, opened 2020-09-04
Support multi-arch buildx docker images

xUnholy:
Signed-off-by: Michael Fornaro <20387402+xUnholy@users.noreply.github.com>

# Description
- Updating Dockerfile to ensure the go binary is compiled for the target architecture platform.
- Adding a test workflow for pushing a release `-multi-arch` suffix image that doesn't break the existing AMD64 image variant.

# TODO
- Discuss other resources that need to be multi-arch in Fluxv2
- Phase `-multi-arch` variant to replace existing.

**EDIT: Just noticed #68 which I think we can slightly improve to use `buildx` and tweak the Dockerfile to accommodate additional architectures in the future on top of arm/amd

### DIFF
--- a/.github/workflows/release-multi-arch.yml
+++ b/.github/workflows/release-multi-arch.yml
@@ -1,0 +1,38 @@
+name: release
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Get version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: crazy-max/ghaction-docker-buildx@v1
+        with:
+          buildx-version: latest
+
+      - name: Login to GitHub Docker Registry
+        run: echo "${DOCKERHUB_PASSWORD}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
+        env:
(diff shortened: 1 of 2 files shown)


stefanprodan:
@xUnholy we are moving away from Docker Hub. See https://github.com/fluxcd/toolkit/issues/194

stefanprodan:
In #68 we use the new Docker action that's based on QEMU and Docker buildx. GHCR doesn't support multi-arch so we have created an arm64 image. We have no plans to support other arches besides amd64/arm64.

xUnholy:
@stefanprodan any idea when GHCR roadmap includes multi-arch support? 

quay also supports multi-arch images and doesn't impose the limit restrictions that docker hub has introduced.

In either regard, I can update this PR to move to GHCR using the QEMU and Docker buildx approach, or shall we close this and just let the PR you raised minutes after move forward? Perfect timing really!

stefanprodan:
We had Flux on Quay and they banned us for to many pulls, Flux is close to 50M pulls per year, I'm not considering going back to Quay like ever lol. Yes please close this PR.

xUnholy:
Closing this PR and will consider reopening a PR when GHCR supposes multi arch images.

The Dockerfile changes could also be considered as overall improvements, even though only reducing layers in the build stage still good practice.

xUnholy:
@stefanprodan GHCR does not support multi-arch, this would be a good feature to investigate as it means there would be no need for the --arch CLI command and it will pull the corresponding image natively, especially nice that there is no need to append image names with arch EG. `-arm64` or `-amd64`.

stefanprodan:
We want to track downloads of amd64 and arm64 separately so the current registry setup will not change anytime soon.

PR #68, opened 2020-09-04
Publish amd64/arm64 images to GHCR

stefanprodan:
Ref: fluxcd/toolkit#194


squaremo (approved):
Looks good, subject to same comments as for https://github.com/fluxcd/source-controller/pull/130

PR #73, opened 2020-09-09
Use pkg/runtime for annotation and predicate

squaremo:
The predicate which filters update events for generation change or the reconcileAt annotation has been factored out into fluxcd/pkg/runtime. This PR removes near-duplicate implementation of the predicate and duplicate definition of the annotation name, and refers to those from fluxcd/pkg/runtime.


hiddeco (approved):
Thank you Michael 🍓

PR #75, opened 2020-09-10
Refactor status API

seaneagan:
Refactor status API
    
    This removes:
    
    - Installed, Upgraded, RolledBack, and Uninstalled status conditions
      since they did not represent current state, but rather actions
      taken, which are already recorded by events.
    - status.observedStateReconciled since it solved the problem of
      remembering past release (install/upgrade/test) success, but not
      past release failures, after other subsequent failures such as
      dependency failures, k8s API failures, etc.
    
    This adds:
    
    - Remediated status condition which records whether the release is
      currently in a remediated state. It is used to prevent release retries
      after remediation failures. We were previously not doing this for
      rollback failures.
    - Released status condition which records whether the current state
      has been successfully released (install/upgrade/test). This is used to
      remember the last release attempt status, regardless of any subsequent
      other failures such as dependency failures, k8s API failures, etc.
    
    This renames:
    
    - Tested > TestSuccess status condition, for forward compatibility
      with interval based helm tests.

xRef: https://github.com/fluxcd/toolkit/discussions/179

### DIFF
--- a/api/v2alpha1/condition_types.go
+++ b/api/v2alpha1/condition_types.go
@@ -51,20 +51,14 @@ const (
 	// ReadyCondition represents the fact that the HelmRelease has been successfully reconciled.
 	ReadyCondition string = "Ready"
 
-	// InstalledCondition represents the fact that the HelmRelease has been successfully installed.
-	InstalledCondition string = "Installed"
+	// ReleasedCondition represents the fact that the HelmRelease has been successfully released.
+	ReleasedCondition string = "Released"
 
-	// UpgradedCondition represents the fact that the HelmRelease has been successfully upgraded.
-	UpgradedCondition string = "Upgraded"
+	// TestSuccessCondition represents the fact that the tests for the HelmRelease are succeeding.
+	TestSuccessCondition string = "TestSuccess"
 
-	// TestedCondition represents the fact that the HelmRelease has been successfully tested.
-	TestedCondition string = "Tested"
-
-	// RolledBackCondition represents the fact that the HelmRelease has been successfully rolled back.
-	RolledBackCondition string = "RolledBack"
-
-	// UninstalledCondition represents the fact that the HelmRelease has been successfully uninstalled.
-	UninstalledCondition string = "Uninstalled"
+	// RemediatedCondition represents the fact that the HelmRelease has been successfully remediated.
+	RemediatedCondition string = "Remediated"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


api/v2alpha1/condition_types.go
```diff
@@ -51,20 +51,14 @@ const (
...
-	// InstalledCondition represents the fact that the HelmRelease has been successfully installed.
-	InstalledCondition string = "Installed"
-
-	// UpgradedCondition represents the fact that the HelmRelease has been successfully upgraded.
-	UpgradedCondition string = "Upgraded"
+	// ReleasedCondition represents the fact that the HelmRelease has been successfully released.
+	ReleasedCondition string = "Released"
 
 	// TestedCondition represents the fact that the HelmRelease has been successfully tested.
 	TestedCondition string = "Tested"
```
hiddeco, line 58: Wondering if something like

```suggestion
	TestedCondition string = "Healthy"
```

would provide a better context to what the condition represents, as `Tested==False` means something entirely different to me on a first glance than what it actually represents.

seaneagan: `TestSucceeded`?

hiddeco: Depends on what it will be used for, if we implement test runs on an interval in a next iteration, the `Healthy` will make more sense I think.

seaneagan: I think the only real use case for this condition is when tests are ignored, since otherwise the `Ready` condition should suffice as it would contain `TestFailed` as the reason. In this case the release may be healthy but the tests are just flaky. I could also imagine a future feature where either helm itself (e.g. a new `helm status` or `helm wait` command) or the helm-controller implement continuous status checking of the resources in a release. In which case we'd want to distinguish this condition from that feature. So we may want to make it clear that this is specific to helm tests.  

I do like the idea of using tense for the interval based test support. Some options:

1. `TestsHealthy`
2. `TestsPassing`
3. `TestsSucceeding`

hiddeco: I think 2 is a nice balance between "human readable" and "machine readable".

seaneagan: Went with `TestSuccess` for now to be consistent with the `TestSucceeded` condition reason, which is consistent with e.g. `InstallSucceeded`, `UpgradeSucceeded`, , `RollbackSucceeded`, `UninstallSucceeded`.

controllers/helmrelease_controller.go
```diff
@@ -330,48 +330,65 @@ func (r *HelmReleaseReconciler) release(ctx context.Context, log logr.Logger, hr
...
-		testErr = r.handleHelmActionResult(&hr, revision, testErr, "test", v2.TestedCondition, v2.TestSucceededReason, v2.TestFailedReason)
-		// Propagate any test error if not marked ignored.
-		if testErr != nil && !remediation.MustIgnoreTestFailures(hr.Spec.GetTest().IgnoreFailures) {
-			err = testErr
+	// If there is a new release revision...
+	if getReleaseRevision(rel) > releaseRevision {
+		// Ensure release is not marked remediated.
+		v2.DeleteHelmReleaseCondition(&hr, v2.RemediatedCondition)
+
+		// If new relrease revision is successful and tests are enabled, run them.
```
hiddeco, line 381:
```suggestion
		// If new release revision is successful and tests are enabled, run them.
```

hiddeco (approved):
Small typo, other than that LGTM :1st_place_medal:

PR #86, opened 2020-09-21
Publish ARM v7/v8 multi-arch image to GHCR

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/246


hiddeco (approved):
Thank you Stefan 🍓

PR #88, opened 2020-09-22
Support optional values references

hiddeco:
This commit adds support for optional values references, as discussions
have brought to light that there are some valid use cases in which
having this option is desirable. For example to allow a user to extend
behaviour at a later date with their own repository without modifying
the HelmRelease object.

When a values reference is marked as optional, not found errors for the
value reference are ignored, but any ValuesKey, TargetPath or transient
error will still result in a reconciliation failure.

Ref: https://github.com/fluxcd/helm-controller/issues/42#issuecomment-693620588

### DIFF
--- a/api/v2alpha1/reference_types.go
+++ b/api/v2alpha1/reference_types.go
@@ -67,6 +67,12 @@ type ValuesReference struct {
 	// Defaults to 'None', which results in the values getting merged at the root.
 	// +optional
 	TargetPath string `json:"targetPath,omitempty"`
+
+	// Optional marks this ValuesReference as optional. When set, a not found
+	// error for the values reference is ignored, but any ValuesKey, TargetPath or
+	// transient error will still result in a reconciliation failure.
+	// +optional
+	Optional bool `json:"optional,omitempty"`
 }
 
 // GetValuesKey returns the defined ValuesKey, or the default ('values.yaml').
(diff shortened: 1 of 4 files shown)


api/v2alpha1/reference_types.go
```diff
@@ -67,6 +67,12 @@ type ValuesReference struct {
 	// Defaults to 'None', which results in the values getting merged at the root.
 	// +optional
 	TargetPath string `json:"targetPath,omitempty"`
+
+	// Optional marks this ValuesReference as optional. When set, a not found
+	// error for the values reference is ignored, but any ValuesKey, TargetPath or
+	// transient errors will still result in a remediation failure.
+	// +optional
+	Optional bool `json:"optional,omitemptya"`
```
stefanprodan, line 75:
```suggestion
	Optional bool `json:"optional, omitempty"`
```

stefanprodan (approved):

controllers/helmrelease_controller.go
```diff
@@ -525,6 +525,9 @@ func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRel
 			var resource corev1.ConfigMap
 			if err := r.Get(ctx, namespacedName, &resource); err != nil {
 				if apierrors.IsNotFound(err) {
+					if v.Optional {
+						continue
```
stefanprodan, line 529: I think we should log on debug that the configmap/secret wasn't found.

PR #92, opened 2020-09-24
Factor out bits and pieces of controller

hiddeco:
Goal of this PR is to factor out logical bits of the controller into packages, to increase code readability and improve testing capabilities.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"crypto/sha1"
 	"errors"
 	"fmt"
 	"io"
@@ -56,6 +55,7 @@ import (
 
 	v2 "github.com/fluxcd/helm-controller/api/v2alpha1"
 	"github.com/fluxcd/helm-controller/internal/runner"
+	"github.com/fluxcd/helm-controller/internal/util"
 )
 
 // HelmReleaseReconciler reconciles a HelmRelease object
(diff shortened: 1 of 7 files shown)


controllers/helmrelease_controller.go
```diff
@@ -267,14 +265,6 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
 }
 
 func (r *HelmReleaseReconciler) release(ctx context.Context, log logr.Logger, hr v2.HelmRelease, source sourcev1.Source, values chartutil.Values) (v2.HelmRelease, error) {
-	// Acquire lock
```
hiddeco, line 282: @seaneagan can you please double check if my observations are right?

seaneagan: Yes it seems controller-runtime ensures only 1 concurrent reconcile per object and `observedGeneration` etc get updated synchronously inside each reconciliation, so makes sense to me.

seaneagan (changes requested):

hiddeco:
@seaneagan current state of the PR is an improvement, but I still have a couple of thoughts / problems I would like to share with you:

1. I think the `helmChartFromTemplate` and `helmChartRequiresUpdate` deserve their own place outside the controller, just have not found a logical place for it (yet).
2. The `handleHelmActionResult` is not really elegant, and I am wondering if we can somehow improve it. Maybe by incorporating some of the logic into the `runner`?
3. I would like to get the `reconcileRelease` method into a state where we can perform tests with it using mocks, to confirm all state determination / decision making / actions add up.
4. `composeValues` makes inefficient API calls at the moment if the `HelmRelease` contains multiple references to the same resource (but with e.g. different paths).
5. `updateStatus` seems an odd wrapper, or an odd place for the wrapper. I have however hopes that by solving 2. this will be magically resolved too :sparkles:

seaneagan:
>     5. `updateStatus` seems an odd wrapper, or an odd place for the wrapper

The only purpose for it is to ensure `status.lastObservedTime` is set. This field provides some value, but doesn't solve all the problems it was intended to, see https://github.com/fluxcd/toolkit/discussions/217. Should we remove it for the v1beta1 release?

stefanprodan (approved):
LGTM

PR #93, opened 2020-09-25
Documentation improvements

seaneagan:
This attempts to work towards the `extend documentation around managing Helm releases` task mentioned in https://github.com/fluxcd/toolkit/issues/238:

The thought here is to have a leveled documentation, increasing detail in each level, and avoiding too much duplication between levels:

1. https://toolkit.fluxcd.io/guides/helmreleases
2. https://toolkit.fluxcd.io/components/helm/helmreleases/
3. https://toolkit.fluxcd.io/components/helm/api/

This:

* copies some of the content about HelmRelease CRD options from 1. to 2. with the idea being to remove or reduce it from 1. and replace with a more prominent link to 2. (as a follow up in the toolkit repo).
* adds more documentation to 2.
* removes some of the copy/pasted golang documentation from 2. since it already exists in 3. The status condition types / reasons are kept (though moved under the status section) since those are not currently included in 3. This could be potentially be resolved in the future by https://github.com/ahmetb/gen-crd-api-reference-docs/issues/21

### DIFF
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -1,11 +1,11 @@
 # Helm Controller
 
-The Helm Controller is a Kubernetes controller, allowing one to declaratively manage Helm chart
-releases with Kubernetes manifests.
+The Helm Controller is a [Kubernetes operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/),
+allowing one to declaratively manage Helm chart releases with Kubernetes manifests.
 
 ## Motivation
 
-The main goal is to provide an automated controller that can perform Helm actions (e.g.
+The main goal is to provide an automated operator that can perform Helm actions (e.g.
 install, upgrade, uninstall, rollback, test) and continuously reconcile the state of Helm releases.
 
 When provisioning a new cluster, one may wish to install Helm releases in a specific order, for
@@ -17,7 +17,7 @@ and this chart needs to be installed first.
 When dealing with an incident, one may wish to suspend the reconciliation of some Helm releases,
 without having to stop the reconciler and affect the whole cluster.
 
-When operating a cluster, different teams may wish to receive notification about the status of
+When operating a cluster, different teams may wish to receive notifications about the status of
 their Helm releases. For example, the on-call team would receive alerts about all failures in
 the prod namespace, while the frontend team may wish to be alerted when a new version of the
 frontend chart was released, no matter the namespace.
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -1,12 +1,12 @@
 # Helm Controller
 
-The Helm Controller is a Kubernetes operator, allowing one to declaratively manage Helm chart
+The Helm Controller is a Kubernetes controller, allowing one to declaratively manage Helm chart
 releases with Kubernetes manifests.
 
 ## Motivation
 
-The main goal is to provide an automated operator that can perform Helm actions (e.g.
-install, upgrade, rollback, test) and continuously reconcile the state of Helm releases.
+The main goal is to provide an automated controller that can perform Helm actions (e.g.
+install, upgrade, uninstall, rollback, test) and continuously reconcile the state of Helm releases.
 
 When provisioning a new cluster, one may wish to install Helm releases in a specific order, for
 example because one relies on a service mesh admission controller managed by a `HelmRelease` and
@@ -34,7 +34,7 @@ actions that should be (conditionally) executed. Based on this the reconciler:
 - fetches the available chart artifact
 - performs a Helm install or upgrade action if needed
 - performs a Helm test action if enabled
-- performs a reconciliation strategy as configured (rollback, uninstall) if any Helm action failed
+- performs a reconciliation strategy (rollback, uninstall) and retries as configured if any Helm action failed
 
 The controller that runs these Helm actions relies on [source-controller](https://github.com/fluxcd/source-controller)
 for providing the Helm charts from Helm repositories or any other source that source-controller
--- a/docs/spec/v2alpha1/helmreleases.md
+++ b/docs/spec/v2alpha1/helmreleases.md
@@ -446,16 +446,17 @@ type HelmReleaseStatus struct {
 
 ```go
 const (
-	// ReleasedCondition represents the fact that the HelmRelease has been
-	// successfully released.
+	// ReleasedCondition represents the status of the last release attempt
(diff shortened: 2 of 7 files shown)


docs/spec/README.md
```diff
@@ -1,12 +1,12 @@
 # Helm Controller
 
-The Helm Controller is a Kubernetes operator, allowing one to declaratively manage Helm chart
+The Helm Controller is a Kubernetes controller, allowing one to declaratively manage Helm chart
```
hiddeco, line 3:
```suggestion
The Helm Controller is a [Kubernetes operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/), allowing one to declaratively manage Helm chart
```

The helm-controller is a controller but also an operator (all operators are controllers but not all controllers are operators), what makes the helm-controller an operator is the fact that it has a single-app focus (Helm).

See: https://kubernetes.io/docs/concepts/extend-kubernetes/operator/

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -1,504 +1,125 @@
...
+## Helm release placement
 
-A **helmrelease** object defines the source of a Helm chart by referencing an object
-managed by [source-controller](https://github.com/fluxcd/source-controller), the interval
-reconciliation should happen at, and a set of options to control the settings of the
-automated Helm actions that are being performed.
+The namespace/name in which to deploy the helm release defaults to the namespace/name of the
+`HelmRelease`. These can be overridden respectively via `spec.targetNamespace` and
+`spec.releaseName`. If `spec.targetNamespace` is set, `spec.releaseName` defaults to
+`<spec.targetNamespace>-<HelmRelease name>`.
```
hiddeco, line 13:
```suggestion
`<spec.targetNamespace>-<metadata.name>`.
```

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -1,504 +1,125 @@
...
+    replicaCount: 2
+  valuesFrom:
+  - kind: ConfigMap
+    name: prod-env-values
+    valuesKey: values-prod.yaml
+  - kind: Secret
+    name: prod-tls-values
+    valuesKey: crt
+	targetPath: tls.crt
+	optional: true
```
hiddeco, line 64:
```suggestion
    targetPath: tls.crt
    optional: true
```

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -1,504 +1,125 @@
...
+action trigger configuration, individual action configuration, and statusing.
 
-## Specification
+## Helm release placement
 
-A **helmrelease** object defines the source of a Helm chart by referencing an object
-managed by [source-controller](https://github.com/fluxcd/source-controller), the interval
-reconciliation should happen at, and a set of options to control the settings of the
-automated Helm actions that are being performed.
+The namespace/name in which to deploy the helm release defaults to the namespace/name of the
```
hiddeco, line 10:
```suggestion
The namespace/name in which to deploy the Helm release defaults to the namespace/name of the
```

As long as we are not directly referring to the binary, the name should be capitalized.

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -1,504 +1,125 @@
...
-[revision](https://github.com/fluxcd/source-controller/blob/master/docs/spec/v1alpha1/common.md#source-status) 
-changes, it generates a Kubernetes event that triggers a new release.
-
-Supported source types:
+## Reconciliation
 
-* [HelmRepository](https://github.com/fluxcd/source-controller/blob/master/docs/spec/v1alpha1/helmrepositories.md)
-* [GitRepository](https://github.com/fluxcd/source-controller/blob/master/docs/spec/v1alpha1/gitrepositories.md)
-* [Bucket](https://github.com/fluxcd/source-controller/blob/master/docs/spec/v1alpha1/buckets.md)
+If a helm release with the matching namespace/name is not found it will be installed, otherwise
```
hiddeco, line 93:
```suggestion
If a Helm release with the matching namespace/name is not found it will be installed, otherwise
```

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -1,504 +1,125 @@
...
+it will be upgraded.
 
-## Reconciliation
+The timeout for any individual Kubernetes operation (like Jobs for hooks) during the performance
+of helm actions can be configured via `spec.timeout` and can be overridden per action
```
hiddeco, line 97:
```suggestion
of Helm actions can be configured via `spec.timeout` and can be overridden per action
```

docs/spec/README.md
```diff
@@ -1,12 +1,12 @@
...
 releases with Kubernetes manifests.
 
 ## Motivation
 
-The main goal is to provide an automated operator that can perform Helm actions (e.g.
-install, upgrade, rollback, test) and continuously reconcile the state of Helm releases.
+The main goal is to provide an automated controller that can perform Helm actions (e.g.
```
hiddeco, line 8: I think the usage of controller is OK here, as "an automated controller that can perform Helm actions" is part of what makes it an operator.

hiddeco (commented):
Thanks for putting time into this :kiwi_fruit:

I have to think for a bit about removing the structures from the spec docs, but do have some comments in the meantime.

docs/spec/README.md
seaneagan: done

hiddeco:
> I have to think for a bit about removing the structures from the spec docs, but do have some comments in the meantime.

I discussed this a bit with @stefanprodan, and we both think the in-documentation code blocks should stay. Reason for this is that once we get out of the "rapid development cycle", new proposals should be made against the spec, and this becomes hard without them being available. We are currently working on more governance (including documentation) for the Flux project, and I will make sure this process will be documented somewhere (cc: @scottrigby).

I also think this means that the documentation within the spec that is added in this PR should be rewritten to something that describes desired application behaviour, instead of being explanatory to a user. But this can be taken care of as a follow up action, including moving bits and pieces to the `HelmRelease` guide (which targets users).

I took the liberty to rebase the PR and undo the code removals from the spec so this can be merged in the meantime, as we are trying to make new controller releases today and would like to have this included.

stefanprodan (approved):
LGTM

GG @seaneagan and @hiddeco

PR #94, opened 2020-09-28
Record last handled reconcile at annotation

hiddeco:
This makes it possible for e.g. the GOTK CLI to observe if the
controller has handled the resource since the manual reconciliation
request was made. It replaces the `LastObservedTime` status field,
as this was prone to time skew issues and does not offer much additional
value over the timestamps of the conditions.

Ref: https://github.com/fluxcd/toolkit/discussions/217


stefanprodan (approved):
LGTM

seaneagan (approved):
There may be some edge cases still (quick succession reconcileAt updates), but I think this makes it easier to determine if a reconcileAt request was handled.

PR #95, opened 2020-09-28
Fix logging for next run delay

seaneagan:
We were logging the spec interval duration, which was incorrect:

1. On failures, which use exponential backoff
2. On dependency not ready, which uses a separately defined static interval.

This changes to log result.RequeueAfter directly when set.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -132,10 +132,11 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	}
 
 	// Log reconciliation duration
-	log.Info(fmt.Sprintf("reconcilation finished in %s, next run in %s",
-		time.Now().Sub(start).String(),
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

stefanprodan:
@seaneagan can you please do this change for all the toolkit controllers?

seaneagan:
> @seaneagan can you please do this change for all the toolkit controllers?

@stefanprodan It looks like the source-controller only logs this message on success in which case the log is correct. The kustomize-controller logs this message on success and some reconcile errors and [attempts](https://github.com/fluxcd/kustomize-controller/blob/e9e9719c041f625e2c4be8bae5c55ab8a44e57d5/controllers/kustomization_controller.go#L192) to requeue after the spec interval in this case, but also returns an error which causes exponential backoff instead. I think we'll want to resolve https://github.com/fluxcd/toolkit/discussions/164 to know exactly what to do here. Do we want to log this message only on success or always? If the latter, we may want some sort of wrapper which looks at the result and does this.

PR #96, opened 2020-09-29
Tweak Helm release namespace/name API

seaneagan:


hiddeco:
> TargetNamespace made more sense for Helm 2 when the release manifests were stored in a separate (target) namespace from the release metadata. In Helm 3 they are one in the same.

There is a catch to this, the `TargetNamespace` defines where the release is deployed to, but **does not** modify the namespace of the storage which will **always** be set to the namespace of the `HelmRelease`. It is thus also not on par with the `--namespace` flag `helm` has, as this will set the namespace of _both_ the release and the storage.

The reason of the above behaviour is because the feature is intended to be used to deploy something to another namespace while hiding the Helm storage from e.g. a tenant.

> This conditional prefixing seems confusing and also doesn't seem to accomplish the presumptive purpose of avoiding release conflicts.

Given the above, it reduces the chance of unexpected conflicts, but is indeed not watertight.

seaneagan:
> > TargetNamespace made more sense for Helm 2 when the release manifests were stored in a separate (target) namespace from the release metadata. In Helm 3 they are one in the same.
> 
> There is a catch to this, the `TargetNamespace` defines where the release is deployed to, but **does not** modify the namespace of the storage which will **always** be set to the namespace of the `HelmRelease`. It is thus also not on par with the `--namespace` flag `helm` has, as this will set the namespace of _both_ the release and the storage.

I missed that, thanks. Does this mean that it would be impossible to do an e.g. `helm upgrade|rollback|uninstall|test` on a Helm release deployed using `HelmRelease` with a different `TargetNamespace` than its namespace, even for an e.g. admin user, since `helm` doesn't seem to expose separate storage / target namespaces?

> > This conditional prefixing seems confusing and also doesn't seem to accomplish the presumptive purpose of avoiding release conflicts.
> 
> Given the above, it reduces the chance of unexpected conflicts (from a tenant point of view), but is indeed not watertight.

If `HelmRelease`s are all stored in the same namespace (e.g. gotk-system or helm-system), their names (and thus default release names) should be unique already, so I guess this is about `HelmRelease`s from different namespaces with the same targetNamespace?

hiddeco:
> I missed that, thanks. Does this mean that it would be impossible to do an e.g. `helm upgrade|rollback|uninstall|test` on a Helm release deployed using `HelmRelease` with a different `TargetNamespace` than its namespace, even for an e.g. admin user, since `helm` doesn't seem to expose separate storage / target namespaces?

That's correct and one of the limitations, I need to start a debate about this upstream to make it e.g. configurable using an environment variable, as they can be configured separately without any issues (proven by our current implementation).

Note that this currently actually does not provide much value, but will in the future when I have tackled the "drift detection" problem. (Halfway there on a local branch, just need to be able to take mutating webhooks into account so that they don't result in spurious upgrades and/or false flags, which should be possible by doing a server-side dry-run apply).

> If HelmReleases are all stored in the same namespace (e.g. gotk-system or helm-system), their names (and thus default release names) should be unique already, so I guess this is about HelmReleases from different namespaces with the same targetNamespace?

It is about a model where the tenant is also creating `HelmRelease` resources _within_ that namespace, and creates one with the same name as the one outside the namespace by accident.

seaneagan:
Thanks for the clarifications! Closing.

gtracer:
I don't quite understand this comment "https://helm.sh/docs/faq/#release-names-are-now-scoped-to-the-namespace" 

@hiddeco could you help me understand why do we need to abstract / hide the helmStorage?

PR #99, opened 2020-09-30
Promote the helm.toolkit.fluxcd.io API to v2beta1

hiddeco:
This PR promotes the `helm.toolkit.fluxcd.io` API to `v2beta1` and removes support for `v2alpha1`.

Going forward, changes to the API will be accompanied by a conversion mechanism. With this release the API becomes more stable but while in beta phase there are no guarantees about backwards compatibility between beta releases.

Ref: https://github.com/fluxcd/toolkit/issues/238


stefanprodan (approved):
LGTM

PR #105, opened 2020-10-05
Prevent possible future nil pointer dereference

hiddeco:
As the existing logic relied on other conditions that caused the
`released` to never be nil, but this may not be true for future
iterations, which makes this a case of better safe than sorry.


seaneagan (approved):
LGTM, good catch

PR #106, opened 2020-10-05
Set default branch to main

hiddeco:
Ref: https://github.com/fluxcd/toolkit/issues/152

### DIFF
--- a/docs/internal/release.md
+++ b/docs/internal/release.md
@@ -2,12 +2,13 @@
 
 To release a new version the following steps should be followed:
 
-1. Create a new branch from `master` i.e. `release-<next semver>`. This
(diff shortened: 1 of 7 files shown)


docs/internal/release.md
```diff
@@ -2,12 +2,12 @@
 
 To release a new version the following steps should be followed:
 
-1. Create a new branch from `master` i.e. `release-<next semver>`. This
```
stefanprodan, line 5: Add this on top please:

```
1. Create a `api/<next semver>` tag and push it to remote.
```

stefanprodan (changes requested):

stefanprodan (approved):

PR #108, opened 2020-10-09
Add note about issues with health checks and helm

phillebaba:


docs/spec/v2beta1/helmreleases.md
```diff
@@ -634,6 +634,10 @@ The timeout for any individual Kubernetes operation (like Jobs for hooks) during
 of Helm actions can be configured via `spec.timeout` and can be overridden per action
 via `spec.<action>.timeout`.
 
+> :warning: **If you are using health checks for HelmRelease**: Helm charts containing a single replica will not behave as expected.
```
hiddeco, line 637: Within the context of the helm-controller, this is not about health checks but about the fact that the wait behaviour does not work as expected (as it does not wait on Helm charts with a single replica).

seaneagan, line 637: A single replica alone will not trigger this issue, needs to also have `maxUnavailble` of 1, which by deafult it's 25% which if replicas is 1, this rounds down to 0, and we don't see the issue. Since this is a long-standing issue with Helm itself, and we cannot enumerate all such issues, do we want to mention it here? Maybe it could be just a Q&A item in the [toolkit discussions](https://github.com/fluxcd/toolkit/discussions)?

phillebaba: Sure that may be better. Do you want to do it and I will close this PR.

seaneagan: https://github.com/fluxcd/toolkit/discussions/338

I'm not opposed to adding the warning here too if we start seeing users report it.

seaneagan: Maybe we should also try to work around this issue in the podinfo chart so users don't hit this when playing around with examples

stefanprodan: Yes let's fix it in podinfo

PR #109, opened 2020-10-13
Implement Prometheus instrumentation

stefanprodan:
This PRs adds Prometheus instrumentation for the HelmRelease reconciliation duration and ready status.

Ref: https://github.com/fluxcd/toolkit/issues/329


controllers/helmrelease_controller.go
```diff
@@ -131,6 +134,9 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		return ctrl.Result{Requeue: true}, updateStatusErr
 	}
 
+	// Record ready status
+	r.recordReadiness(hr, false)
```
seaneagan, line 138: We should also call this on L165 and L322 (after the calls to `v2.HelmReleaseProgressing()`)

seaneagan (changes requested):

controllers/helmrelease_controller.go
stefanprodan: Done!

hiddeco (approved):

seaneagan (approved):

PR #110, opened 2020-10-13
Add namespace to API docs examples

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/319


hiddeco (approved):
Thank you Stefan :medal_sports:

PR #114, opened 2020-10-14
Declare and assign result Values before composing

### LINKED ISSUE #113: Helm release fails with 'unable to parse key: assignment to entry in nil map'
We are trying to deploy linkerd with the following `HelmRelease`

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: linkerd
  namespace: default
spec:
  chart:
    spec:
      chart: linkerd2
      sourceRef:
        kind: HelmRepository
        name: linkerd
  interval: 1m0s
  releaseName: linkerd
  targetNamespace: default
  valuesFrom:
    - kind: Secret
      name: linkerd-ca
      valuesKey: ca.crt
      targetPath: global.identityTrustAnchorsPEM
    - kind: Secret
      name: linkerd-issuer
      valuesKey: issuer.crt
      targetPath: identity.issuer.tls.crtPEM
    - kind: Secret
      name: linkerd-issuer
      valuesKey: issuer.key
      targetPath: identity.issuer.tls.keyPEM
```

We get the following error message when we run `gotk get hr -n default`

```
unable to merge value from key 'ca.crt' in Secret 'default/linkerd-ca' into target path 'global.identityTrustAnchorsPEM': unable to parse key: assignment to entry in nil map
```

We have noticed that the error is happening at this line: https://github.com/fluxcd/helm-controller/blob/6402c7d775ca6905de84e0a192ddc4e7fa42f8a5/controllers/helmrelease_controller.go#L554
as `result` is still nil.

hiddeco:
As `ParseInto` expects the destination to be initialized, which led
to an `unable to parse key: assignment to entry in nil map` error in
configurations where the first `ValuesReference` had a `targetPath`
defined.

Fixes #113

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -462,7 +462,7 @@ func (r *HelmReleaseReconciler) checkDependencies(hr v2.HelmRelease) error {
 // and merges them as defined. Referenced resources are only retrieved once
 // to ensure a single version is taken into account during the merge.
 func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRelease) (chartutil.Values, error) {
-	var result chartutil.Values
+	result := chartutil.Values{}
 
 	configMaps := make(map[string]*corev1.ConfigMap)
 	secrets := make(map[string]*corev1.Secret)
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

PR #124, opened 2020-10-22
Use reconcile annotation helpers

squaremo:
Per https://github.com/fluxcd/pkg/issues/39


hiddeco (approved):
Thanks @squaremo :cake:

PR #127, opened 2020-10-26
Split controller logic into scoped methods

hiddeco:
The summary of changes (till date) is as follows:

* Use pointers for API objects; this makes it easy to transfer state
  between methods and limitizes the chance that we run in to
  `the object has been modified` errors from the API server.
* Split reconcile actions into dedicated methods; they are independent
  and able to make decisions based on the current state of the object
  they retrieve. This also makes it much easier to test their logic,
  as an addition to having end-to-end tests.
* Only use `Client#Update` when we actually update things that are not
  the status object (those are at present: de(registering) finalizers).
  For all other status updates, use `Client#Status#Update`.
* Aggregate multiple returned errors using
  `k8s.io/apimachinery/pkg/util/errors`.
* Do not touch a suspended object, as it suspended. If we want to make
  the suspendend state surface in e.g. `kubectl`, we should simply get
  the value from the `spec` field instead of pushing a `Ready==False`
  condition.


api/v2beta1/helmrelease_types.go
```diff
@@ -241,9 +241,9 @@ type Remediation interface {
 	MustIgnoreTestFailures(bool) bool
 	MustRemediateLastFailure() bool
 	GetStrategy() RemediationStrategy
-	GetFailureCount(hr HelmRelease) int64
+	GetFailureCount(hr *HelmRelease) int64
```
squaremo, line 244: I don't see much point in changing these from values to pointers; at least with values, it signals that the method does not mutate the value.

squaremo: (Since I can't put a comment on an unchanged line, it goes here: the Remediation interface should really be defined in the package that consumes arguments of that type, not here.)

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,232 @@
...
+func (r *HelmReleaseReconciler) reconcileRemediation(ctx context.Context, run *runner.Runner, hr *v2.HelmRelease) (ctrl.Result, error) {
+	// If the release did not explicitly fail, we should not remediate
+	released := meta.GetCondition(hr.Status.Conditions, v2.ReleasedCondition)
+	if released == nil || released.Status != v1.ConditionFalse {
+		return ctrl.Result{}, nil
+	}
+
+	// Get the right remediation strategy based on the failure count,
+	// looking back, this would have been easier if we did not decide
+	// to flatten our conditions (https://github.com/fluxcd/helm-controller/pull/75)
```
hiddeco, line 184: @seaneagan please have a look at this in the context of the code in this PR (please note, it is an early stage draft and some details are not 100% right yet, I plan to resolve those while working on tests in Go).

controllers/helmrelease_controller.go
```diff
@@ -252,189 +82,135 @@ func (r *HelmReleaseReconciler) SetupWithManager(mgr ctrl.Manager, opts HelmRele
...
-	hr.Status.LastReleaseRevision = util.ReleaseRevision(rel)
-
-	if err != nil {
-		reason := meta.ReconciliationFailedReason
-		var cerr *ConditionError
-		if errors.As(err, &cerr) {
-			reason = cerr.Reason
+	result := ctrl.Result{}
+	var errs []error
+	for _, step := range steps {
```
squaremo, line 198: I like this approach; is it not the case that you want to halt at the first error, though?

hiddeco: No, the idea is that there may be introduced logic later that e.g. settles things if your controller crashed mid-run. If you simply run through all steps while requiring that the methods called self-determine if they actually should do anything, this always gives them a chance every reconciliation run to correct state.

seaneagan: So would this be for example retries of helm test/rollback/uninstall actions?

controllers/helmrelease_controller.go
```diff
@@ -252,189 +82,135 @@ func (r *HelmReleaseReconciler) SetupWithManager(mgr ctrl.Manager, opts HelmRele
...
 		}
-
-		// Fail if install retries are exhausted.
-		if hr.Spec.GetInstall().GetRemediation().RetriesExhausted(hr) {
-			err = fmt.Errorf("install retries exhausted")
-			return v2.HelmReleaseNotReady(hr, released.Reason, released.Message), err
+		// Record observed generation
+		hr.Status.ObservedGeneration = hr.Generation
+		if retErr != nil {
+			v2.HelmReleaseNotReady(hr, meta.ReconciliationFailedReason, err.Error())
```
seaneagan, line 162: Are we bubbling up the reasons/messages (e.g. install/upgrade/test failures) to the Ready condition yet like we do today?

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,232 @@
...
+		if err := r.Client.Status().Update(ctx, hr); err != nil {
+			return ctrl.Result{}, err
+		}
+		r.recordReadiness(hr, false)
+	}
+
+	// If the previous release was successful, return
+	if released := meta.GetCondition(hr.Status.Conditions, v2.ReleasedCondition); released != nil {
+		if released.Status == v1.ConditionTrue {
+			return ctrl.Result{}, nil
```
seaneagan, line 74: We should "remember" the previous Ready reason/message here, which is stored in the Released condition, since it may have been lost in e.g. a transient dependency not ready reconciliation.

seaneagan (changes requested):

controllers/helmrelease_controller.go
hiddeco: Exactly, or running the tests without having made a release, etc.

controllers/helmrelease_controller.go
hiddeco: This is bubbling up the error(s) returned by the `reconcile*` methods to the `Ready` message.

controllers/helmrelease_controller_release.go
hiddeco: What is not reflected here, is that I originally started with also splitting the `reconcileRelease` into a `reconcileInstall` and `reconcileUpgrade`. Not knowing exactly what happened during a previous reconciliation (even while taking into account the failure counts), made this impossible.

It is true that you would still need to cobble together conditions at times, but it would made state determination much easier, and the code much more readable. At current it feels like we are at times following tiny breadcrumbs to determine state, which is also a quite brittle approach in my opinion.

controllers/helmrelease_controller.go
seaneagan: Ok, so just not bubbling up the `Reason`s yet. That would be very useful to keep as well if we can. For example anything that uses the `kstatus` library which consumes only the `Ready` condition, will then be able to surface the underlying `Reason` which provides a quick "at a glance" diagnosis for the end user, and some level of machine readability of the status.

controllers/helmrelease_controller_release.go
seaneagan: @hiddeco The `Released` condition already exposes `InstallFailed` or `UpgradeFailed` reasons. Currently it also includes `TestFailed` in its determination, but if we simply remove keep the helm test result encapsulated in the `TestSuccess` condition only, and record the `Released` condition directly after the install/upgrade (i.e. in `reconcileRelease` / `reconcileInstall` / `reconcileUpgrade`), then this should provide the explicit previous install/upgrade failure detection we are looking for here?

controllers/helmrelease_controller.go
seaneagan: With the approach of having separate steps for `reconcileRelease`(or `reconcileUpgrade`/`reconcileInstall`)/`reconcileTest`, how do we accomplish incrementing the install/upgrade failure counts, which depends on knowing whether there was a *new* failure in either the install/upgrade or a resulting (non-ignored) test? I guess we could increment the failure count in `reconcileRelease`(or `reconcileUpgrade`/`reconcileInstall`) upon failure, and then in `reconcileTest` determine the reason(s) why a helm test needs to be run:

1. successful install (maps to `installFailures`)
2. successful upgrade (maps to `upgradeFailures`)
3. https://github.com/fluxcd/flux2/discussions/103

and then upon test failure, only increment a failure count which is mapped to by the calculated reason(s) for having running the test (1. or 2.)

seaneagan: Nevermind, looks like this is already implemented 👍

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,232 @@
...
+		if released.Reason == v2.UpgradeSucceededReason {
+			remediation = hr.Spec.GetUpgrade().GetRemediation()
+		}
+
+		if remediation.MustIgnoreTestFailures(hr.Spec.GetTest().IgnoreFailures) {
+			return ctrl.Result{}, nil
+		}
+
+		remediation.IncrementFailureCount(hr)
+		v2.SetHelmReleaseCondition(hr, v2.ReleasedCondition, v1.ConditionFalse, v2.TestFailedReason, "")
```
seaneagan, line 166: ...so just remove this line, and the `Released` condition becomes a reliable (barring status update failures) indicator of the last install/upgrade attempt against the current state.

hiddeco:
@seaneagan thanks for your additional round of review and thoughts. I did not have a chance last week to continue the work on this due to other priorities, but am hoping to revive the state of this PR (later) this week :sun_with_face:

controllers/helmrelease_controller_release.go
seaneagan: I guess though if we remove the test result from the Released condition, and later implement https://github.com/fluxcd/flux2/discussions/103, then the initial test result could get overwritten by subsequent interval-based tests. So then in `reconcileRemediation` we wouldn't be able to distinguish between an initial test failure and an interval based one without adding some additional distinguishing bit of status. Perhaps this could be simply using an alternate `reason` in the `TestSuccess` condition, e.g.  `IntervalTest<Succeeded|Failed>` (vs just `Test<Succeeded|Failed>` for the initial test).

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	rel, err = doRelease(*hr, loadedChart, values)
+	if err != nil {
+		v2.SetHelmReleaseCondition(hr, v2.ReleasedCondition, v1.ConditionFalse, failureReason, "")
+		remediation.IncrementFailureCount(hr)
+		r.event(*hr, hr.Status.LastAttemptedRevision, events.EventSeverityError,
+			fmt.Sprintf("Helm release failed: %s", err.Error()))
+		return ctrl.Result{}, err
+	}
+
```
seaneagan, line 124: I think we still need to account for deleting the `RemediatedCondition` here:

https://github.com/fluxcd/helm-controller/blob/9b0daa031acfad05e148f0fae76cdaa454b4a658/controllers/helmrelease_controller.go#L393-L396

PR #131, opened 2020-10-28
Move dedicated watcher to in-controller watcher

hiddeco:
This prevents the resources from getting annotated, and instead uses
the `handler.EnqueueRequestsFromMapFunc` to queue requests based on
changes to the source objects.


config/rbac/role.yaml
```diff
@@ -41,9 +41,3 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
-  - source.toolkit.fluxcd.io
-  resources:
-  - helmcharts/status
-  verbs:
-  - get
```
stefanprodan, line 49: Don't we need to check the status of a HelmChart?

hiddeco: I think `GET`ting the status object is already covered by being able to access the resource itself, I made the same change to the kustomize-controller which had no problems with it: https://github.com/fluxcd/kustomize-controller/pull/153/files#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7

stefanprodan: kustomize and helm controllers have no problem with it due to the cluster-admin binding, but I assure you that on OpenShift it will error out without the cluster-admin binding, I had to add the status to RBAC in Flagger, that's how I know :)

stefanprodan: OpenShift also needs `/finalizers`

stefanprodan (approved):

PR #133, opened 2020-10-28
Compare artifact <> HelmRelease in enqueuer

hiddeco:
The reason for this is the `EnqueueRequestsFromMapFunc` calling the
enqueuer for _both_ the old and the new object, and we only want to act
on the ones that contain a revision different from the one that we have
recorded in the status object of the `HelmRelease`.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -96,8 +96,8 @@ func (r *HelmReleaseReconciler) SetupWithManager(mgr ctrl.Manager, opts HelmRele
 		For(&v2.HelmRelease{}, builder.WithPredicates(predicates.ChangePredicate{})).
 		Watches(
 			&source.Kind{Type: &sourcev1.HelmChart{}},
-			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.helmReleasesForHelmChart)},
-			builder.WithPredicates(HelmChartRevisionChangePredicate{}),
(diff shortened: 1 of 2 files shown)


squaremo:
What I see is:

 - the watch on `HelmChart` only processes updates if the revision changed in the update
 - each event coming from that watch will enqueue the HelmRelease objects that both 1. use the HelmChart (they are selected using the index); and 2. report a last attempted revision different to that of the HelmChart.

That seems reasonable to me. Are there changes to a HelmChart aside from the revision that would matter? (My guess is probably not -- if something else changed, it would only matter if it resulted in a new revision being current.)

controllers/helmrelease_controller.go
```diff
@@ -704,7 +704,15 @@ func (r *HelmReleaseReconciler) handleHelmActionResult(hr *v2.HelmRelease, revis
 	}
 }
 
-func (r *HelmReleaseReconciler) helmReleasesForHelmChart(obj handler.MapObject) []reconcile.Request {
+func (r *HelmReleaseReconciler) enqueueForHelmChartArtifactChange(obj handler.MapObject) []reconcile.Request {
+	hc, ok := obj.Object.(*sourcev1.HelmChart)
+	if !ok {
+		panic(fmt.Sprintf("Expected a HelmChart, got %T", hc))
```
stefanprodan, line 710: Can we return an error instead of crashing the process?

hiddeco: The reason I use a panic here is because it should never happen. If it happens, something is wired wrong (by us) and logging this does not make this obvious.

stefanprodan (approved):

PR #128, opened 2020-10-26
Upgrade Helm to v3.4.0

hiddeco:
This Helm version includes a fix to properly log the output of Pod
status phases for (test) jobs.

Fixes #11

PR #136, opened 2020-10-30
Helm dependency update to v3.4.0

relu:
Note that this also comes with k8s 1.19 which brings in klog v2, hence the update on the logr and respectively zapr (indirect, just for compatibility reasons) packages.

Can supersede #128


hiddeco:
Closing, as #128 was in good shape and just required a rebase and a line addition.

PR #148, opened 2020-11-18
Refactor to adopt k8s standardized Condition type

hiddeco:
Changes the condition type to the one introduced in K8S 1.19, including
the newly introduced helpers in place of the old pkg/apis/meta types.

Depends on https://github.com/fluxcd/source-controller/pull/195


controllers/helmrelease_controller.go
```diff
@@ -693,12 +699,22 @@ func (r *HelmReleaseReconciler) garbageCollectHelmRelease(logger logr.Logger, hr
...
 		return &ConditionError{Reason: failedReason, Err: errors.New(msg)}
 	} else {
 		msg := fmt.Sprintf("Helm %s succeeded", action)
-		v2.SetHelmReleaseCondition(hr, condition, corev1.ConditionTrue, succeededReason, msg)
+		apimeta.SetStatusCondition(&hr.Status.Conditions, metav1.Condition{
+			Type:    condition,
+			Status:  metav1.ConditionTrue,
+			Reason:  succeededReason,
+			Message: msg,
+		})
```
seaneagan, line 717: Don't we want to use https://github.com/fluxcd/pkg/pull/49 here so that we will get the observed generation handling once it is implemented?

controllers/helmrelease_controller.go
```diff
@@ -402,8 +403,13 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, log logr.L
...
 			if testErr != nil && !remediation.MustIgnoreTestFailures(hr.Spec.GetTest().IgnoreFailures) {
-				testsPassing := meta.GetCondition(hr.Status.Conditions, v2.TestSuccessCondition)
-				v2.SetHelmReleaseCondition(&hr, v2.ReleasedCondition, corev1.ConditionFalse, testsPassing.Reason, testsPassing.Message)
+				testsPassing := apimeta.FindStatusCondition(hr.Status.Conditions, v2.TestSuccessCondition)
+				apimeta.SetStatusCondition(&hr.Status.Conditions, metav1.Condition{
+					Type:    v2.ReleasedCondition,
+					Status:  metav1.ConditionFalse,
+					Reason:  testsPassing.Reason,
+					Message: testsPassing.Message,
+				})
```
seaneagan, line 412: Don't we want to use https://github.com/fluxcd/pkg/pull/49 here so that we will get the observed generation handling once it is implemented?

seaneagan (dismissed):

stefanprodan (approved):

PR #156, opened 2020-11-24
Requeue after interval on source not found errors

stefanprodan:
In case the artifact watcher doesn't receive the chart update event, we should requeue based on the specfied interval to avoid a forever stuck HelmRelease.

Fix: #153


hiddeco (approved):
Thank you @stefanprodan :coffee:

PR #159, opened 2020-11-25
Cleanup end-to-end test objects after each run

hiddeco:
- Remove end-to-end test objects after a successful run
- Deduplicate test sources


stefanprodan (approved):
LGTM 💯

PR #162, opened 2020-11-25
Do not mark suspended resource as not ready

hiddeco:
If a resource is marked as suspended we should not perform any
reconciliation action for the resource at all.

This includes updating the status of the resource to a
`Ready==False` state, as this is not an accurate representation of
the resource's state: we have been told to no longer look at it
and with that our last observation freezes in time.

Resource deletions are the sole exception to actions (not) being
performed, as we are duty-bound to remove our finalizer from the
resource to make a deletion succeed.

To make the suspend state still visible to the user, a column has
been added to the Custom Resource Definition which prints the
`.spec.suspend` of every HelmRelease.

```
❯ kubectl get helmrelease
NAME          READY   STATUS                             SUSPENDED   AGE
podinfo       True    Release reconciliation succeeded   true        22m
podinfo-git   True    Release reconciliation succeeded   false       22m
```

Ref: https://github.com/fluxcd/flux2/issues/507

### DIFF
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -51,8 +51,10 @@ type HelmReleaseSpec struct {
 
 	// Suspend tells the controller to suspend reconciliation for this HelmRelease,
 	// it does not apply to already started reconciliations. Defaults to false.
+	// +kubebuilder:default:=false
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM 💯

PR #166, opened 2020-11-26
Refactor release reconciliation

hiddeco:


controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+	// The state has not changed and the release is not in a failed state.
+	case rls != nil && remediation.GetFailureCount(*hr) == 0:
+		v2.HelmReleaseReady(hr, rls.Info.Description)
+		return ctrl.Result{RequeueAfter: hr.Spec.Interval.Duration}, nil
+	// We have exhausted our retries.
+	case remediation.RetriesExhausted(*hr):
+		v2.HelmReleaseNotReady(hr, meta.ReconciliationFailedReason, "exhausted release retries")
+		return ctrl.Result{RequeueAfter: hr.Spec.Interval.Duration}, nil
+	// Our previous reconciliation attempt failed, skip release to retry.
+	case hr.Status.LastSuccessfulReleaseRevision > 0 && hr.Status.LastReleaseRevision != hr.Status.LastSuccessfulReleaseRevision:
```
seaneagan, line 123: The `LastSuccessfulReleaseRevision` seems to be managed by `reconcileTest`, why not check for the `Released==True` here instead, which is managed by this method (`reconcileRelease`)?

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+		return ctrl.Result{RequeueAfter: hr.Spec.Interval.Duration}, nil
+	}
```
seaneagan, line 125: Do we still have anything here to ensure that failures were successfully remediated before retrying? I wasn't able to find that.

api/v2beta1/helmrelease_types.go
```diff
@@ -653,10 +653,14 @@ type HelmReleaseStatus struct {
...
 	LastAttemptedValuesChecksum string `json:"lastAttemptedValuesChecksum,omitempty"`
 
-	// LastReleaseRevision is the revision of the last successful Helm release.
+	// LastReleaseRevision is the revision of the last performed Helm release.
 	// +optional
 	LastReleaseRevision int `json:"lastReleaseRevision,omitempty"`
 
+	// LastSuccessfulReleaseRevision is the revision of the last successful Helm release.
+	// +optional
+	LastSuccessfulReleaseRevision int `json:"lastSuccessfulReleaseRevision,omitempty"`
```
seaneagan, line 662: In this PR, this only seems to be used to tell if the `LastReleaseReivision` was successful, which seems like it could also be derived from the `Released` and `TestSuccess` conditions (and whether tests are enabled). Are there other reasons to add this field, such as for manual rollbacks?

seaneagan (changes requested):
Added some initial feedback and questions after a first pass.

api/v2beta1/helmrelease_types.go
hiddeco: It is used to roll back to a revision we have made ourselves, see the change in the `Rollback` method.

controllers/helmrelease_controller_release.go
hiddeco: While working on this, it became cumbersome to derive state from the conditions as a failure may or may not result in a new revision, and precise reflection of this is hard.

By opting to keep track of revision numbers instead in combination with observations of the Helm storage writes, we can re-observe based on what is recorded in the Helm storage for those revision numbers, instead of having the blindly trust our observations from the past.

controllers/helmrelease_controller_release.go
hiddeco: The comment should be: `// Our previous remediation attempt failed, skip release to retry`, which tries to settle the remediation before continuing with another attempt.

internal/storage/observer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Name returns the name of the driver.
+func (o *Observer) Name() string {
+	return ObserverDriverName
+}
+
+// Get returns the release named by key or returns ErrReleaseNotFound.
+func (o *Observer) Get(key string) (*release.Release, error) {
+	return o.Driver.Get(key)
```
seaneagan, line 63: Do we want to set `o.release` here?

internal/storage/observer.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if hasNewerRevision(o.release, rls) {
+		o.release = rls
+	}
+	return nil
+}
+
+// Delete deletes a release or returns driver.ErrReleaseNotFound.
+func (o *Observer) Delete(key string) (*release.Release, error) {
+	return o.Driver.Delete(key)
```
seaneagan, line 104: Do we need to lock here as well?

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+	r.event(hr, hr.Status.LastAttemptedRevision, events.EventSeverityInfo, run.GetLastPersistedRelease().Info.Description)
+
+	return ctrl.Result{RequeueAfter: hr.Spec.Interval.Duration}, nil
+}
+
+func (r *HelmReleaseReconciler) reconcileTest(ctx context.Context, log logr.Logger, run *runner.Runner, hr *v2.HelmRelease) (ctrl.Result, error) {
+	// If this release was already marked as successful,
+	// we have nothing to do.
+	if hr.Status.LastReleaseRevision == hr.Status.LastSuccessfulReleaseRevision {
+		return ctrl.Result{}, nil
```
seaneagan, line 171: Shouldn't we always be setting `requeueAfter: hr.Spec.Interval.Duration` here and elsewhere when there is nothing to do, to ensure we don't end up using whatever the default value is? Or alternatively should we override to that duration after running each step if there were no errors and the requeueAfter is zero?

seaneagan (changes requested):

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+	values, err := r.composeValues(ctx, log, hr)
+	if err != nil {
+		v2.HelmReleaseNotReady(hr, v2.InitFailedReason, err.Error())
+		r.event(hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	// Observe the last release. If this fails, we likely encountered a
+	// transient error and should return it to requeue a reconciliation.
+	rls, err := run.ObserveLastRelease(hr)
```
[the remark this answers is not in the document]
hiddeco: `ObserveLastRelease` retrieves the last release from the Helm storage, and this may return a manually made release, the "cached" version is returned by `GetLastPersistedRelease`.

internal/storage/observer.go
hiddeco: We do not at this moment, as it does not touch the cached object.

internal/storage/observer.go
hiddeco: Likely not, because we only want to observe writes we make ourselves. If we would set `o.release` here, the object could be overwritten / "poisoned" during a Helm action with a manually made release, as most actions fetch the current release from storage to use as a template for the next release they are about to store.

controllers/helmrelease_controller.go
```diff
@@ -142,262 +138,104 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
-	hr.Status.LastReleaseRevision = util.ReleaseRevision(rel)
-
-	if err != nil {
-		reason := meta.ReconciliationFailedReason
-		var cerr *ConditionError
-		if errors.As(err, &cerr) {
-			reason = cerr.Reason
+		if len(errs) > 0 {
+			continue
 		}
```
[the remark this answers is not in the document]
hiddeco: It skips the calculation of the `LowestNonZeroResult` as an error always results in a requeue.

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+		hr.Status.LastReleaseRevision = util.ReleaseRevision(run.GetLastPersistedRelease())
+		hr.Status.LastSuccessfulReleaseRevision = hr.Status.LastReleaseRevision
+
+		// If we did not keep our Helm release history, we are now
+		// back to square one.
+		if !hr.Spec.GetUninstall().KeepHistory {
+			hr.Status.LastReleaseRevision = 0
+			hr.Status.LastSuccessfulReleaseRevision = 0
+		}
+	}
```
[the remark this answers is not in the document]
hiddeco: It is not, as Helm always writes to the storage on uninstall but truncates the history one step later if `KeepHistory` is set to `false`. This means that we always have a copy of a deleted release, while the release may no longer be present in the Helm storage.

internal/storage/observer.go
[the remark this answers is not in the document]
hiddeco: Likely not, because that would result in garbage collection of the cached release when the max number of revisions is reached if we do not take the `key` somehow into account. Sadly, [the knowledge about the key format is private](https://github.com/helm/helm/blob/master/pkg/storage/storage.go#L247), which is why I was hesitant in utilizing it.

What we may be able to do is to store the `o.release` as a key/value pair, that may also be an enabler for https://github.com/fluxcd/helm-controller/pull/166#discussion_r533522218, if we add an additional `deleted` flag to keep the release.

controllers/helmrelease_controller_release.go
hiddeco: The latter has my preference I think.

internal/runner/runner.go
```diff
@@ -97,12 +110,13 @@ func (r *Runner) Rollback(hr v2.HelmRelease) error {
 	rollback.Force = hr.Spec.GetRollback().Force
 	rollback.Recreate = hr.Spec.GetRollback().Recreate
 	rollback.CleanupOnFail = hr.Spec.GetRollback().CleanupOnFail
+	rollback.Version = hr.Status.LastSuccessfulReleaseRevision
```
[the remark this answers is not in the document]
hiddeco: > There is an existing release, such as one previously managed by the Helm Operator, and Helm controller takes it over.

If `rollback.Version` is set to `0`, Helm will roll it back to the previous revision.

> User manually fixes a helm release, perhaps via a suspend/resume flow.

Given this creates "revision drift", it would always trigger a new release and I would expect this to succeed because of the fixes performed by the user (and later reflected to the chart and/or `HelmRelease` before resuming).

> Also, if a previous reconciliation had a failed upgrade which was not rolled back, then do we really want to rollback to a previous release which may have been made a long time ago and may be out of date?

I am hesitant about this because of the following:

1. We have little knowledge what the contents of the previous revision are, or if it has been tampered with.
2. We have little knowledge about the state of previous revision before the current revision was created (was it e.g. `StatusFailed`?), as any "previous" revision gets a `StatusSuperseded` from Helm.
3. Our "revision bookkeeping" model would make less sense, as we would detect mutations but allow rollbacks to the previous state if an upgrade on top of this detection fails. Given we then mark the revision of the rolled back release as the new revision, the controller basically would verify state it did not create itself.

controllers/helmrelease_controller_release.go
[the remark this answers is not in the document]
seaneagan: Ah, I got confused between `ObserveLastRelease` and `GetLastObservedRelease` whose names make it seem like they are more closely related than they are.

seaneagan (commented):
Still going through this, but if I understand correctly (didn't actually test), the install/upgrade retry exhaustion now no longer applies to test and remediation actions. So we can now have infinite retries of remediation actions (uninstall/rollback), which could cause the release revisions to rev to infinity, in the case of repeatedly failed rollbacks. Also when a release is not remediated, either intentionally (`remediateLastFailure=false`), or has a failed remediation, this can lead to infinite test retries as well, since `LastSuccessfulReleaseRevision != LastReleaseRevision`, which could lead to silently ignoring test failures when a subsequent one succeeds, which may be ok but one probably wants a way to prevent that. It also may lead to problems if one is not using the appropriate [hook deletion policies](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies) on their helm tests. It also seems like we could be testing an old release we rolled back to instead of the current desired state.

hiddeco:
@seaneagan noted, and thank you for taking the time.

I am open to suggestions for improvements and/or alternative approaches, as the issues you have brought up do seem valid (but preventable) at this moment (although it is late here, and I need to reweigh them again tomorrow with fresh energy).

Given the goal of this PR is to improve testability, my item on the list for this PR is to create Go tests for the reconciler actions so that we can properly test and cover the uncertainties you describe.

First step to achieve this is to untangle some of the `Runner` code to make it possible to inject Helm's [fake `kube` client](https://github.com/helm/helm/tree/master/pkg/kube/fake) and their memory based storage driver, after that it should be fairly easy to mock state and confirm the reconciliation outcome.

hiddeco:
> the install/upgrade retry exhaustion now no longer applies to test and remediation actions

It still does? https://github.com/fluxcd/helm-controller/blob/release-refactor/controllers/helmrelease_controller_release.go#L276-L280

> Also when a release is not remediated, either intentionally (`remediateLastFailure=false`), or has a failed remediation, this can lead to infinite test retries as well, since `LastSuccessfulReleaseRevision != LastReleaseRevision`, which could lead to silently ignoring test failures when a subsequent one succeeds, which may be ok but one probably wants a way to prevent that.

I think this actually all captured by the state observations we do on the release object itself? https://github.com/fluxcd/helm-controller/blob/release-refactor/controllers/helmrelease_controller_release.go#L183-L200

seaneagan:
> > the install/upgrade retry exhaustion now no longer applies to test and remediation actions
> 
> It still does? https://github.com/fluxcd/helm-controller/blob/release-refactor/controllers/helmrelease_controller_release.go#L276-L280

Sorry, not sure how I missed that. One scenario I'm still not sure about, if we have say 5 upgrade retries enabled, and the first upgrade fails, and then the rollback repeatedly fails, the upgrade retries will thus be skipped, and thus the release revision will continue to rev to infinity for each new failed rollback. I assume solving this would mean adding a retry limit (whether configurable or not) and failure count status field for rollbacks. I don't think uninstall has the same issue since a failure to uninstall shouldn't rev the release revision, although perhaps it makes sense to stop retrying at some point anyways, not sure.

> > Also when a release is not remediated, either intentionally (`remediateLastFailure=false`), or has a failed remediation, this can lead to infinite test retries as well, since `LastSuccessfulReleaseRevision != LastReleaseRevision`, which could lead to silently ignoring test failures when a subsequent one succeeds, which may be ok but one probably wants a way to prevent that.
> 
> I think this actually all captured by the state observations we do on the release object itself? https://github.com/fluxcd/helm-controller/blob/release-refactor/controllers/helmrelease_controller_release.go#L183-L200

Missed that as well, thanks.

controllers/helmrelease_controller_release.go
```diff
@@ -0,0 +1,426 @@
...
+		r.event(hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, msg)
+		return ctrl.Result{}, err
+	}
+
+	// Determine the release strategy based on the existence of a
+	// release in the Helm storage.
+	makeRelease, remediation := run.Install, hr.Spec.GetInstall().GetRemediation()
+	successReason, failureReason := v2.InstallSucceededReason, v2.InstallFailedReason
+	if rls != nil {
+		makeRelease, remediation = run.Upgrade, hr.Spec.GetUpgrade().GetRemediation()
```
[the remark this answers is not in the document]
hiddeco: If we change it to `rls != nil && hr.Status.LastSuccessfulReleaseRevision > 0`, the install strategy is picked for `HelmRelease` resources that are "taking over" a resource, and may result in an accidental wipe of all release data if that operation fails.

Issue #130, opened 2020-10-27
Allow configurable use of --create-namespace


tommasopozzetti:
I am re-opening here the issue fluxcd/helm-operator#488 since helm-operator is now in maintenance mode and this issue was getting some attention.
I do believe having the option to explicitly control the `--create-namespace` flag via the HelmRelease could be extremely beneficial  for certain use-cases, such as the one mentioned by @voron:
```
Helm2 has "create namespace" by default, helm3 re-invented this feature starting from 3.2.
I'm here because I want to create all HelmRelease in single namespace and helm-operator with clusterRole will create all the required namespaces based on targetNamespace then.
```

With this new setup and the new HelmRelease CRD, implementing this should be as easy as adding a field to the [Install](https://github.com/fluxcd/helm-controller/blob/v0.1.3/api/v2beta1/helmrelease_types.go#L233) configuration which in turn would control whether this flag is set or not when calling helm install commands (defaulting to false).

hiddeco:
Support has been added in `v0.4.1` of the helm-controller, and made available in `v0.4.2` of `flux`.

PR #168, opened 2020-11-27
Support Helm's NS creation for TargetNamespace

hiddeco:
This adds support for creating the target release namespace if it is not
present which can be be useful in certain scenarios.

Note that when the release is uninstalled, the namespace is not removed
and remains on the cluster, and managing your namespace _outside_ of
the HelmRelease is advised.

Fixes: #130


rk185151:
This feature is useful. Please review this PR. Thanks!

stefanprodan (approved):
LGTM

PR #189, opened 2021-01-12
Upgrade source-controller to v0.6.0

hiddeco:
This commit upgrades both the `source-controller` API package and the
runtime dependency to `v0.6.0`.


stefanprodan (approved):
🚀

PR #191, opened 2021-01-13
Remove mergo version replacement

seaneagan:
As of #183 this is no longer needed.


hiddeco (approved):
Your commit needs a sign-off, but LGTM :+1:

PR #200, opened 2021-01-25
Release v0.6.1

hiddeco:


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.6.1
+
+**Release date:** 2021-01-25
+
+This prerelease adds support for configuring the namespace of the
+Helm storage by defining a `StorageNamespace` in the `HelmRelease`
```
stefanprodan, line 8: There is no mention of this in https://github.com/fluxcd/helm-controller/blob/main/docs/spec/v2beta1/helmreleases.md

stefanprodan: > The Helm storage is stored on the remote cluster in a namespace that equals to the namespace of the HelmRelease, the release itself is made in either this namespace, or the configured spec.targetNamespace. In any case, both are expected to exist.

This also needs to change.

stefanprodan (approved):

Issue #104, opened 2020-10-02
Support Helm PostRender Functionality


RichiCoder1:
# Describe the feature
It might be a little harder given the toolkit's new structure, but it'd be great to see support for [Helm's Post Render](https://helm.sh/docs/topics/advanced/#post-rendering) functionality.

This allows for customization of out-of-the box charts without having to manually untar and manipulate the base chart.

## Alternatives

Probably an advanced feature, but being able to chain Helm and Kustomize controllers would be another way to attack this.

kferrone:
Maybe just allow a Flux specific post render. I suggested to Crossplane to include a simple patch method to their Helm controller, they agreed. I suggested to implement a simple post render json patch. Then implementation would look just like how kustomize does patches, something like:

```yaml
patches:
- patch: |-
    - op: replace
      path: /some/existing/path
      value: new value
  target:
    kind: MyKind
    labelSelector: "env=dev"
```

Then you don't need to deal with the chaos of all the different ways a post render may look and this basically covers most possible use cases.

hiddeco:
This is a great future request and slipped through the cracks while I collected "highly upvoted issues" from the Helm Operator to determine the feature set for the helm-controller. It had already resurfaced in my mind due to discussions with @scottrigby, but had not been recorded in an issue yet, so thank you for doing just that. :sunflower:

Your proposal @kferrone, would serve use-cases where you want to apply a simple patch to the manifests rendered by Helm very well. This _may_ be sufficient to most users, but due to the fact that the feature in Helm is still relatively unknown[[1]](https://twitter.com/hiddeco/status/1303301666998018049), there is a lack of real world (non-declarative) examples to confirm this.

However, if this turns out to be limiting, we will eventually hear about it from our users and can then think about extending the support for other post-renderers (one way that has crossed my mind is by utilizing jobs with user defined images, but note that keeping track of jobs is not a light introduction), as long as we design the API properly.

seaneagan:
[...]

PR #202, opened 2021-02-01
Support for Kustomize based PostRenderer

alex-berger:
This PR adds support for Kustomize based [Post Renderer](https://helm.sh/docs/topics/advanced/#post-rendering) to the HelmRelease resource. This PR is a first draft, to gather feedback from the project maintainers and while it compiles is has not yet been well tested. 

**Goals:**
- Add simple built-in Post Renderer capability which supports as subset of Kustomize ([patchesStrategicMerge](https://kubectl.docs.kubernetes.io/references/kustomize/patchesstrategicmerge/), [patchesJson6902](https://kubectl.docs.kubernetes.io/references/kustomize/patchesjson6902/) and [images](https://kubectl.docs.kubernetes.io/references/kustomize/images/)) and which will work with the default helm-controller image.
- Add minimal framework to support more Post Renderers in future.

**Non Goals:**
- Supporting arbitrary non-built-in Post Renderers which might require special image with external tools included.

# Example Usage

```yaml
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: metrics-server
  namespace: kube-system
spec:
  interval: 1m
  chart:
    spec:
      chart: metrics-server
      version: "5.3.4"
      sourceRef:
        kind: HelmRepository
        name: bitnami
        namespace: kube-system
      interval: 1m
  postRenderers:
  - kustomize: 
       patchesStrategicMerge:
       -  kind: Deployment
          apiVersion: apps/v1
          metadata:
            name: metrics-server
            namespace: kube-system
          spec:
            template:
              spec:
                tolerations:
                - key: "workload-type"
                  operator: "Equal"
                  value: "cluster-services"
                  effect: "NoSchedule"
```

# Motivation

As SRE / DevOps engineer I often have the need to patch Helm Charts from third-parties, either by forking them or by contributing to the upstream via pull requests. Both approaches are cumbersome and time intensive and often cause project delays and errors. Having simple built-in Helm Post Renderer support in `helm-controller` will allows me to patch Helm Releases immediately without having to fork a Helm Chart or waiting for upstream contributions to be integrated. 

This PR also (partially) addresses #104


hiddeco (commented):
Thank you for putting time into this, as it has been sought after :tulip: 

To not directly import a lot of new functionalities (and half of Kustomize), and to encourage people to continue to make changes to charts upstream, I think it would be best if [we limit the set of functionalities to what was discussed in the issue](https://github.com/fluxcd/helm-controller/issues/104#issuecomment-703362247).

This would cover most people their base needs, and we can then start a discussion around any further expansions.

api/v2beta1/helmrelease_types.go
```diff
@@ -32,6 +33,37 @@ import (
 const HelmReleaseKind = "HelmRelease"
 const HelmReleaseFinalizer = "finalizers.fluxcd.io"
 
+// PatcheJSON6902 JSON 6902 patches
+type PatcheJSON6902 struct {
+	// Patch is the YAML content of a patch.
+	Patch []apiextensionsv1.JSON `json:"patch,omitempty" yaml:"patch,omitempty"`
+
+	// Target points to the resources that the patch is applied to
+	Target kustypes.Selector `json:"target,omitempty" yaml:"target,omitempty"`
```
hiddeco, line 42: We can not directly rely on the Kustomize API from within our own API, as it has been unstable at times. The kustomize-controller [works around this by duplicating types](https://github.com/fluxcd/kustomize-controller/blob/89017ca9cce6233b423ea657daff76d2178155bc/api/v1beta1/kustomization_types.go#L126-L139).

alex-berger: OK, I refactored this by duplicating the types.

internal/runner/post_renderer.go
```diff
@@ -0,0 +1,35 @@
+package runner
+
+import (
+	"bytes"
+
+	"helm.sh/helm/v3/pkg/postrender"
+)
+
+// combinedPostRenderer, a collection of Helm PostRenders which are
+// invoked in the order or insertion.
```
hiddeco, line 10:
```suggestion
// combinedPostRenderer, a collection of Helm PostRenders which are
// invoked in the order of insertion.
```

internal/runner/post_renderer.go
```diff
@@ -0,0 +1,35 @@
...
+	return combinedPostRenderer{
+		renderers: make([]postrender.PostRenderer, 0),
+	}
+}
+
+func (c combinedPostRenderer) addRenderer(renderer postrender.PostRenderer) {
+	c.renderers = append(c.renderers, renderer)
+}
+
+// implement PostRenderer interface:
```
hiddeco, line 25: I would simply remove this comment.

internal/runner/post_renderer_kustomize.go
```diff
@@ -0,0 +1,132 @@
...
+		output.Gvk.Version = selector.Gvk.Version
+		output.Name = selector.Name
+		output.Namespace = selector.Namespace
+		output.LabelSelector = selector.LabelSelector
+		output.AnnotationSelector = selector.AnnotationSelector
+	}
+	return
+}
+
+// implement PostRenderer interface:
```
hiddeco, line 74: Same as for the other comment.

internal/runner/post_renderer_kustomize.go
```diff
@@ -0,0 +1,132 @@
...
+	spec *v2.Kustomize
+}
+
+func newPostRendererKustomize(spec *v2.Kustomize) *postRendererKustomize {
+	return &postRendererKustomize{
+		spec: spec,
+	}
+}
+
+func writeFile2(fs filesys.FileSystem, path string, content []byte) error {
```
hiddeco, line 24:
```suggestion
func writeToFile(fs filesys.FileSystem, path string, content []byte) error {
```

api/v2beta1/helmrelease_types.go
```diff
@@ -32,6 +32,37 @@ import (
...
+	// +optional
+	PatchesStrategicMerge []apiextensionsv1.JSON `json:"patchesStrategicMerge,omitempty"`
+	// JSON 6902 patches, defined as inline YAML objects.
+	// +optional
+	PatchesJSON6902 []PatchJSON6902 `json:"patchesJson6902,omitempty"`
+	// Images is a list of (image name, new name, new tag or digest)
+	// for changing image names, tags or digests. This can also be achieved with a
+	// patch, but this operator is simpler to specify.
+	// +optional
+	Images []Image `json:"images,omitempty" yaml:"images,omitempty"`
```
hiddeco, line 56: Can you add newlines between the fields?

api/v2beta1/kustomization_types.go
```diff
@@ -0,0 +1,45 @@
+package v2beta1
```
hiddeco, line 1: Can you please add a copyright header to this file?

hiddeco (commented):
Couple of tiny nitpicks, otherwise looking OK to me. :+1:

It would also be awesome if you could reflect to changes made to the in-code API in the documentation in `docs/spec/v2beta1/helmreleases.md`.

**NB:** you accidentally checked the full `helm-controller` binary into your commit, you probably want to remove this from the commit and `push --force-with-lease` :-)

hiddeco (commented):
Some additional end-to-end tests with base coverage would be very welcome too, to ensure the full picture is wired correctly.

api/go.mod
```diff
@@ -8,4 +8,5 @@ require (
 	k8s.io/apiextensions-apiserver v0.20.2
 	k8s.io/apimachinery v0.20.2
 	sigs.k8s.io/controller-runtime v0.8.0
+	sigs.k8s.io/kustomize/api v0.7.2
```
stefanprodan, line 11: Can we remove this and undo all the 171 additions to the API go.sum?

hiddeco (approved):
Awesome work, and if you smash your commits together I would be happy to merge them! :tada: 

The helm-controller release that is going to include these changes is going to take a little longer, as we want to move some of the API elements introduced here to `fluxcd/pkg` so they can be shared with the kustomize-controller. Work on this should however not take too long. :sunflower:

stefanprodan (approved):

PR #205, opened 2021-02-03
Depend on the Kustomize API from `fluxcd/pkg`

hiddeco:


api/v2beta1/helmrelease_types.go
```diff
@@ -49,13 +41,13 @@ type Kustomize struct {
 
 	// JSON 6902 patches, defined as inline YAML objects.
 	// +optional
-	PatchesJSON6902 []PatchJSON6902 `json:"patchesJson6902,omitempty"`
+	PatchesJSON6902 []kustomize.JSON6902Patch `json:"patchesJson6902,omitempty"`
```
stefanprodan, line 44: We should rename this to `patches` as `patchesJson6902` has been deprecated.

stefanprodan (approved):

PR #206, opened 2021-02-04
Optimize Kustomize post renderer and tests

hiddeco:
* Use constants for APIVersion and Kind in Kustomization config
* Remove redundant JSON marshal of strategic merge patches
* Factor out Kustomization build wrapper and add notes about settings
* Add tests for Kustomize post renderer


stefanprodan (approved):
Nice refactoring 🥇

PR #216, opened 2021-02-14
Allow the controller to be run locally

stefanprodan:
This PR makes it possible to run helm-controller locally for debugging purposes against a test cluster.

### DIFF
--- a/controllers/helmrelease_controller_chart.go
+++ b/controllers/helmrelease_controller_chart.go
@@ -98,7 +98,16 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
 	defer f.Close()
 	defer os.Remove(f.Name())
 
-	res, err := http.Get(source.GetArtifact().URL)
+	url := source.GetArtifact().URL
+	if hostname := os.Getenv("SOURCE_CONTROLLER_LOCALHOST"); hostname != "" {
+		url = fmt.Sprintf("http://%s/%s/%s/%s/latest.tar.gz",
+			hostname,
+			strings.ToLower(source.Kind),
+			source.GetNamespace(),
+			source.GetName())
+	}
+
+	res, err := http.Get(url)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


stefanprodan:
A more suitable label would be `area/DX` 😺

controllers/helmrelease_controller_chart.go
```diff
@@ -98,7 +98,16 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
 	defer f.Close()
 	defer os.Remove(f.Name())
 
-	res, err := http.Get(source.GetArtifact().URL)
+	url := source.GetArtifact().URL
+	if hostname := os.Getenv("SOURCE_CONTROLLER_LOCALHOST"); hostname != "" {
+		url = fmt.Sprintf("http://%s/%s/%s/%s/latest.tar.gz",
```
hiddeco, line 103: Shouldn’t this address simply be advertised by the resource object if is running locally?

stefanprodan: The whole point of this is that you can run only HC locally.

stefanprodan: If someone whats to contribute to HC, now it needs to compile and run SC locally and that bings lots of pain due to libgit2. With this, you can debug HC locally while all the others controllers run in-cluster. Please see the guide I wrote in this PR.

hiddeco: Check, I think it would be better to mutate the URL by parsing it using `url.Parse` and then modify the hostname. As `latest.tar.gz` (and the rest of the URL format) is an assumption that does not work for charts: https://github.com/fluxcd/source-controller/blob/568b2845c7e06abdb0c4efb365abf91a819225d5/controllers/helmchart_controller.go#L444

stefanprodan: Ah I forgot about that...

hiddeco: That's what reviews are for :-D

hiddeco (approved):
Thank you @stefanprodan :cake:

PR #217, opened 2021-02-15
add a release deployment event when reconciling a release

william-salt-cko:
At the moment, you are not sure if the release has started and you have to wait for events showing its success or failure, or watch the status of k8s resources.
This adds an event notifying us that a release has started, we have found this to be especially useful for our slack messages in our deployment channels.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -322,6 +322,8 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 		}
 	}
 
+	// Send event to mark the beginning release deploy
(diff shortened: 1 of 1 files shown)


hiddeco (commented):
Would it make sense to send separate events for install or upgrade, so that it is clear from the event what action will be run?

william-salt-cko:
@hiddeco yep, can do! Is that ok? I thought i would stick to similar wording as the success event being `Helm upgrade succeeded`.

hiddeco:
@william-salt-cko I think that in e.g. notification messages it is helpful to be notified of the precise action, because seeing an installation while you were expecting an upgrade may help you detect an e.g. miss-configuration faster.

hiddeco (approved):
Thank you @william-salt-cko :tulip:

PR #219, opened 2021-02-15
Capture and expose debug (log) information on release failure

### LINKED ISSUE #218: Capture and expose debug (log) information on wait timeout
If a helm release has `wait` enabled, the default setting, there is a chance of failure due to something obscure and hard to detect which is only logged in debug mode. Like when a service: LoadBalancer is used on a home lab cluster that doesn't enable LB, opting instead for HostNetworking ingress. Helm will wait for the LoadBalancer service to become ready, which it will never do, and the error message will be `context deadline exceeded` or some fairly unhelpful message that is not by itself enough to trace back to the source of the problem. You have to enable debug mode.

This is not strictly a Helm Controller bug, since HC faithfully reproduces the behavior of Helm upstream in this regard, in debug operating mode, before a timeout is reached, Helm fails reconciling with this error (and so does helm-controller):

```
client.go:464: [debug] Looks like there are no changes for Ingress "controller-api-server-ingress-http"
wait.go:53: [debug] beginning wait for 59 resources with timeout of 30s
wait.go:225: [debug] Service does not have load balancer ingress IP address: deis/deis-builder
wait.go:225: [debug] Service does not have load balancer ingress IP address: deis/deis-builder
...
# on like this until "context deadline exceeded"
```

The helpful information is hidden behind the `--debug` gateway, which is possible to access in helm-controller now by setting `--log-level=debug` on the entire helm controller.

So, filed under DX/UX note, I thought wouldn't it be great if with my HelmRelease that fails to reconcile, there was a `debug` flag in the API that I can use on a per-HelmRelease basis, to enable emitting debug entries into (at least the helm-controller logs?)

In thinking this through, it occurs to me now too that perhaps the most useful place to have the debug logs would be in the kubernetes event stream for the HelmRelease, and not just in the helm-controller logs... I have not tried to enable the log-level debug yet, not sure which of these it does right now. But either way, I can see wanting to enable `--debug` logs from Helm on just a particular HelmRelease vs doing so on every helm release that hc reconciles across the entire cluster at once.

Related conversation from CNCF slack: https://cloud-native.slack.com/archives/CLAJ40HV3/p1613417136364700?thread_ts=1613335794.255500&cid=CLAJ40HV3

hiddeco:
- [x] Enable devel flag to get richer debug information from Helm
- [x] Buffer `n` last log lines for the action
- [x] Include the log information in the failure event when the install or upgrade fails

Fixes #218.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -328,14 +328,14 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 		// Fail if install retries are exhausted.
 		if hr.Spec.GetInstall().GetRemediation().RetriesExhausted(hr) {
 			err = fmt.Errorf("install retries exhausted")
-			return v2.HelmReleaseNotReady(hr, released.Reason, released.Message), err
+			return v2.HelmReleaseNotReady(hr, released.Reason, err.Error()), err
 		}
 
 		// Fail if there is a release and upgrade retries are exhausted.
 		// This avoids failing after an upgrade uninstall remediation strategy.
 		if rel != nil && hr.Spec.GetUpgrade().GetRemediation().RetriesExhausted(hr) {
 			err = fmt.Errorf("upgrade retries exhausted")
-			return v2.HelmReleaseNotReady(hr, released.Reason, released.Message), err
+			return v2.HelmReleaseNotReady(hr, released.Reason, err.Error()), err
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


stefanprodan:
@hiddeco may I suggest we buffer the log lines after eliminating duplicates? This way an ingress/lb/volume being stuck will not pollute the event.

stefanprodan:
I would prioritize this to trim down the support burden, users can’t figure out why releases are failing when wait times out.

stefanprodan (approved):
LGTM

Thanks @hiddeco

controllers/helmrelease_controller.go
```diff
@@ -328,14 +328,14 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 		// Fail if install retries are exhausted.
 		if hr.Spec.GetInstall().GetRemediation().RetriesExhausted(hr) {
 			err = fmt.Errorf("install retries exhausted")
-			return v2.HelmReleaseNotReady(hr, released.Reason, released.Message), err
+			return v2.HelmReleaseNotReady(hr, released.Reason, err.Error()), err
```
hiddeco, line 331: This (and the change below) are required because otherwise the long error message will still end up in the `Ready` condition.

It will also make the error more visible to users, as I know from providing support to people that it is a bit hidden at the moment.

PR #228, opened 2021-02-26
Retry with exponential backoff when fetching artifacts

stefanprodan:
This PR implements retries with exponential backoff when fetching artifacts from source-controller. By default, the controller does 10 attempts within a 3.5 minutes window, the number of max retries can be configure using the `--http-retry` cmd arg.

This mitigates the alert spam (`i/o timeout err`) when source-controller becomes unavailable for a short period of time e.g. after an upgrade. 

Ref: https://github.com/fluxcd/flux2/discussions/661 https://github.com/fluxcd/notification-controller/issues/76

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


hiddeco (approved):
:100: thanks @stefanprodan

PR #229, opened 2021-03-03
fix: log messages contain '%s'

makkes:
Before:

```
{"level":"info","ts":"2021-03-03T12:45:41.305Z","logger":"controller.helmrelease","msg":"could not find optional %s '%s'","reconciler group":"helm.toolkit.fluxcd.io","reconciler kind":"HelmRelease","name":"cert-manager","namespace":"kommander","ConfigMap":{"namespace":"kommander","name":"cert-manager-overrides"}}
```

After:

```
{"level":"info","ts":"2021-03-03T13:56:46.365+0100","logger":"controller.helmrelease","msg":"could not find optional ConfigMap 'kommander/cert-manager-overrides'","reconciler group":"helm.toolkit.fluxcd.io","reconciler kind":"HelmRelease","name":"cert-manager","namespace":"kommander"}
```

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -540,7 +540,8 @@ func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRel
 				if err := r.Get(ctx, namespacedName, resource); err != nil {
 					if apierrors.IsNotFound(err) {
 						if v.Optional {
-							(logr.FromContext(ctx)).Info("could not find optional %s '%s'", v.Kind, namespacedName)
+							(logr.FromContext(ctx)).
+								Info(fmt.Sprintf("could not find optional %s '%s'", v.Kind, namespacedName))
 							continue
 						}
 						return nil, fmt.Errorf("could not find %s '%s'", v.Kind, namespacedName)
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @makkes 🏅

controllers/helmrelease_controller.go
```diff
@@ -540,7 +540,8 @@ func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRel
 				if err := r.Get(ctx, namespacedName, resource); err != nil {
 					if apierrors.IsNotFound(err) {
 						if v.Optional {
-							(logr.FromContext(ctx)).Info("could not find optional %s '%s'", v.Kind, namespacedName)
+							(logr.FromContext(ctx)).
+								Info(fmt.Sprintf("could not find optional %s '%s'", v.Kind, namespacedName))
```
jimmidyson, line 544: Wouldn't these all be better as structured logs, something like:
```suggestion
								Info("could not find optional resource", "kind", v.Kind, "name", namespacedName)
```

makkes: I was thinking the same thing but I'm not sure about the policy here. @stefanprodan wdyt?

hiddeco: If we opt for structured logging, we should ensure that the structures are the same across all GOTK controllers.

This is an OK feature request (and probably would be better), but requires coordination.

jimmidyson: Makes sense @hiddeco.

stefanprodan: We do have the reconciled object kind/name/namespace as log labels, I don't think the secret kind/name belongs there as it will overlap with the HelmRelease and users will not be able to tell what's going on.

hiddeco (approved):
Thank you @makkes :medal_sports:

PR #238, opened 2021-03-18
Label release resources with HelmRelease origin

hiddeco:
This commit adds a new post renderer that labels all resources with
`helm.toolkit.fluxcd.io/name` and `helm.toolkit.fluxcd.io/namespace`
so their source of origin can be traced back by e.g. the Flux UI.

The post renderer makes use of the Kustomize API without running
a full Kustomize build, by making directly use of the builtin
`LabelTransformerPlugin` on a `ResMap` that has been constructed
from the bytes of the `bytes.Buffer` given by Helm.

### DIFF
--- a/internal/runner/post_renderer_origin_labels.go
+++ b/internal/runner/post_renderer_origin_labels.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2021 The Flux authors
+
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM 💯 

Thanks @hiddeco

PR #239, opened 2021-03-23
Set leader election deadline to 30s

stefanprodan:
This PR allows fine tuning the leader election with cmd args:

```
--enable-leader-election=true
--leader-election-release-on-cancel=true
--leader-election-lease-duration=35s
--leader-election-renew-deadline=30s
--leader-election-retry-period=5s
```

To make the leader election more tolerant to transient Kubernetes API outages, the renewal deadline was increased from 10s to 30s and the retry interval was bumped to 5 seconds.

Test image:

```
docker.io/stefanprodan/helm-controller:le-config-b74081d
```

Part of: https://github.com/fluxcd/pkg/issues/94

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
 	"time"
 
@@ -40,6 +41,8 @@ import (
 	"github.com/fluxcd/helm-controller/controllers"
 	// +kubebuilder:scaffold:imports
 )
+
+const controllerName = "helm-controller"
 
 var (
 	scheme   = runtime.NewScheme()
@@ -85,7 +88,7 @@ func main() {
 
 	var eventRecorder *events.Recorder
 	if eventsAddr != "" {
-		if er, err := events.NewRecorder(eventsAddr, "helm-controller"); err != nil {
+		if er, err := events.NewRecorder(eventsAddr, controllerName); err != nil {
 			setupLog.Error(err, "unable to create event recorder")
 			os.Exit(1)
 		} else {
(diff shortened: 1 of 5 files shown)


hiddeco (approved):
Tested various disruptions on a Kubernetes v1.20.4 "production" cluster without any issues. Thanks @stefanprodan :100:

main.go
```diff
@@ -105,14 +103,18 @@ func main() {
...
+		Scheme:                        scheme,
+		MetricsBindAddress:            metricsAddr,
+		HealthProbeBindAddress:        healthAddr,
+		Port:                          9443,
+		LeaderElection:                leaderElectionOptions.Enable,
+		LeaderElectionReleaseOnCancel: leaderElectionOptions.ReleaseOnCancel,
+		LeaseDuration:                 &leaderElectionOptions.LeaseDuration,
+		RenewDeadline:                 &leaderElectionOptions.RenewDeadline,
+		RetryPeriod:                   &leaderElectionOptions.RetryPeriod,
+		LeaderElectionID:              "5b6ca942.fluxcd.io",
```
hiddeco, line 115: Something that comes to mind now that we are touching leader election configurations is to change the ID to something more identifying, as the current IDs have been autogenerated by kubebuilder.

[CAPI seems to use lengthy strings](https://github.com/kubernetes-sigs/cluster-api/search?q=LeaderElectionID), but I think `helm-controller.fluxcd.io` would also suffice.

hiddeco: Other option would be `flux-helm-controller-leader-election`.

hiddeco (approved):
Excellent use of a value that was already present :100:

PR #245, opened 2021-04-12
Conform to k8s sample to allow for auto generation of clientset, listers, informers

lwj5:
The API should conform to the directory structure as such https://github.com/kubernetes/sample-controller to allow for automatic code generation.

The generated clientset, listers, and informers should be packaged with the module which can be used in the controller. `controllers` and `internal` folders should be moved into pkg as well. These are not done in this PR.

I'll generate the clientset, listers, and informers and update the build file should this be approved


hiddeco:
We are not following the K8s example, but the way kubebuilder (v2) works. The suggested way of using the object types there is not by creating a specific client set, but rather to use the already exposed types in combination with the dynamic client, by using `AddToScheme`

See: https://github.com/kubernetes-sigs/kubebuilder/issues/1152#issuecomment-584774562 and https://hackernoon.com/platforms-on-k8s-with-golang-watch-any-crd-0v2o3z1q

lwj5:
Ahhh, thanks for the references! Shall read up on the new way.

PR #250, opened 2021-04-19
Initial support for HelmRelease for upgrading CRDs

alex-berger:
This PR adds preliminary support for Upgrading CRDs which are part of Helm Charts managed via `HelmRelease` objects.

See also https://github.com/fluxcd/flux2/discussions/1071 for more information.

## Background & Motivation

Helm still does not provide any built-in solution to the CRD upgrade problem, which is well-known and
[documented](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/):

> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

This limitation of Helm makes GitOps style "day 2 operations" of Kubernetes resources managed by
`HelmRelease` objects very difficult if not to say impossible. Currently, we have to manually
upgrade CRDs from `HelmCharts` referenced by `HelmReleases`, which is cumbersome and needs very 
tight (timing) coordination between the commit to the GitOps repository and the manual CRD 
upgrade on all systems that observe/apply that commit. This apporach is not only very error 
prone it also might cause unnecessary long downtime of services.

Note, most of the Helm Charts that we install and operate are created and maintained by 
3rd parties and are not under our control. Extracting all CRDs from every Helm Chart 
upon each new release and manually applying those CRD resources is, as mentioned above,
very time intensive, error prone and cumbersome.

Our observation is, that most CRD upgrades are non-critical and only evolve an existing CRD
in a backward compatible fashion. Therefore, I am wondering whether it would make sense to
extends the `HelmRelease` resources with an opt-in flag to automatically upgrade CRD objects
(if needed).

### DIFF
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -431,6 +431,19 @@ func (in InstallRemediation) RetriesExhausted(hr HelmRelease) bool {
 	return in.Retries >= 0 && in.GetFailureCount(hr) > int64(in.Retries)
 }
 
+// CRDsUpgradePolicy defines the upgrade approach to use for CRDs when upgrading
+// a HelmRelease.
+type CRDsChangePolicy string
+
+const (
+	// Create CRDs which do not already exist, do not replace already existing CRDs
+	// and keep (do not delete) CRDs which no longer exist in the current release.
+	Create CRDsChangePolicy = "Create"
+	// Create CRDs which do not already exist, Replace already existing CRDs
+	// and keep (do not delete) CRDs which no longer exist in the current release.
+	CreateReplace CRDsChangePolicy = "CreateReplace"
+)
+
 // Upgrade holds the configuration for Helm upgrade actions for this
 // HelmRelease.
 type Upgrade struct {
@@ -473,6 +486,24 @@ type Upgrade struct {
 	// upgrade action when it fails.
 	// +optional
 	CleanupOnFail bool `json:"cleanupOnFail,omitempty"`
+
+	// UpgradeCRDs upgrade CRDs from the Helm Chart's crds directory according
+	// to the CRD upgrade policy provided here. Valid values are `Create` or
+	// `CreateReplace`. If omitted (the default) CRDs
+	// are not upgraded.
+	//
+	// Create: new CRDs are created, existing CRDs are neither updated nor deleted.
+	//
+	// CreateReplace: new CRDs are created, existing CRDs are updated (replaced)
+	// but not deleted.
+	//
+	// By default, CRDs are not applied during Helm upgrade action. With this
+	// option users can opt-in to CRD upgrade, which is not (yet) natively supported by Helm.
+	// https://helm.sh/docs/chart_best_practices/custom_resource_definitions.
+	//
+	// +kubebuilder:validation:Enum=Create;CreateReplace
+	// +optional
+	UpgradeCRDs CRDsChangePolicy `json:"upgradeCRDs,omitempty"`
 }
 
 // GetTimeout returns the configured timeout for the Helm upgrade action, or the
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -335,6 +335,10 @@ type Install struct {
 
 	// SkipCRDs tells the Helm install action to not install any CRDs. By default,
 	// CRDs are installed if not already present.
+	//
+	// Deprecated use CRD policy (`crds`) attribute with value `Skip` instead.
+	//
+	// +deprecated
 	// +optional
 	SkipCRDs bool `json:"skipCRDs,omitempty"`
 
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -338,6 +338,27 @@ type Install struct {
 	// +optional
 	SkipCRDs bool `json:"skipCRDs,omitempty"`
 
+	// CRDs upgrade CRDs from the Helm Chart's crds directory according
+	// to the CRD upgrade policy provided here. Valid values are `Skip`,
+	// `Create` or `CreateReplace`. Default is `Create` and if omitted
+	// CRDs are installed but not updated.
+	//
+	// Skip: do neither install nor replace (update) any CRDs.
+	//
+	// Create: new CRDs are created, existing CRDs are neither updated nor deleted.
+	//
+	// CreateReplace: new CRDs are created, existing CRDs are updated (replaced)
+	// but not deleted.
+	//
+	// By default, CRDs are applied (installed) during Helm install action.
+	// With this option users can opt-in to CRD replace existing CRDs on Helm
+	// install actions, which is not (yet) natively supported by Helm.
+	// https://helm.sh/docs/chart_best_practices/custom_resource_definitions.
+	//
+	// +kubebuilder:validation:Enum=Skip;Create;CreateReplace
+	// +optional
+	CRDs CRDsPolicy `json:"crds,omitempty"`
+
 	// CreateNamespace tells the Helm install action to create the
 	// HelmReleaseSpec.TargetNamespace if it does not exist yet.
 	// On uninstall, the namespace will not be garbage collected.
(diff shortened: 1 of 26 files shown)


internal/runner/runner.go
```diff
@@ -131,14 +143,168 @@ func (r *Runner) Upgrade(hr v2.HelmRelease, chart *chart.Chart, values chartutil
...
+	switch policy {
+	case "":
+		break
+	case v2.Create:
+		break
+	case v2.CreateReplace:
+		break
+	default:
+		return policy, errors.New(
+			fmt.Sprintf("Invalid CRD upgrade policy '%s' defined in field upgradeCRDs, valid values are '%s' or '%s'",
```
hiddeco, line 176:
```suggestion
		return policy, fmt.Errorf("invalid CRD upgrade policy '%s' defined in field upgradeCRDs, valid values are '%s' or '%s'"),
```

docs/spec/v2beta1/helmreleases.md
```diff
@@ -1109,6 +1127,53 @@ spec:
...
+    spec:
+      chart: my-operator
+      version: "1.0.1"
+      sourceRef:
+        kind: HelmRepository
+        name: my-operator-repo
+        namespace: default
+      interval: 1m
+  upgrade:
+	upgradeCRDs: CreateReplace
```
hiddeco, line 1174:
```suggestion
    upgradeCRDs: CreateReplace
```

internal/runner/runner.go
```diff
@@ -131,14 +143,168 @@ func (r *Runner) Upgrade(hr v2.HelmRelease, chart *chart.Chart, values chartutil
...
+		// Read in the resources
+		res, err := cfg.KubeClient.Build(bytes.NewBuffer(obj.File.Data), false)
+		if err != nil {
+			cfg.Log("failed to parse CRDs from %s: %s", obj.Name, err)
+			return errors.New(fmt.Sprintf("failed to parse CRDs from %s: %s", obj.Name, err))
+		}
+		allCrds = append(allCrds, res...)
+	}
+	totalItems := []*resource.Info{}
+	if policy == v2.Create {
```
hiddeco, line 205: Using a `switch` here reads (and expands) better imo.

alex-berger:
Unfortunately, I run out of ideas how to make the e2e test working with pull request. E2e tests work on my feature branches which are real branches that can be referenced by the `GitRepository`, but pull request are not real branches so it fails.

api/v2beta1/helmrelease_types.go
```diff
@@ -463,6 +476,24 @@ type Upgrade struct {
...
+	// CreateReplace: new CRDs are created, existing CRDs are updated (replaced)
+	// but not deleted.
+	//
+	// By default, CRDs are not applied during Helm upgrade action. With this
+	// option users can opt-in to CRD upgrade, which is not (yet) natively supported by Helm.
+	// https://helm.sh/docs/chart_best_practices/custom_resource_definitions.
+	//
+	// +kubebuilder:validation:Enum=Create;CreateReplace
+	// +optional
+	UpgradeCRDs CRDsChangePolicy `json:"upgradeCRDs,omitempty"`
```
hiddeco, line 496: Having looked at the implementation now, I see room for install to support using the `CreateReplace` policy during install, given all logic is there.

However, if we would do this, I think the flags probably should be changed to be included in the action specific configurations (`spec.install` and `spec.upgrade`).

```yaml
spec:
  # Fallback default
  crds: "None"
  install:
    crds: "Create"
  upgrade:
	# Default: "None"
    crds: "CreateReplace"
```

When `install.crds == Create` it is probably best to use the Helm config flag instead of our own logic.

The old configuration flag should be marked deprecated but still taken into account.

What do you think?

hiddeco (commented):
This already looks great, couple of minor comments and a suggestion that I would like to discuss. :medal_sports:

api/v2beta1/helmrelease_types.go
alex-berger: I am not yet sure how to implement this, but let me share the key tenets that I have in mind:

- Keep backward compatibility, this PR should not introduce any breaking change to existing `HelmRelease` objects. This implies that for `install` the default must be `Create` and for `upgrade` the default must be `Skip`.
- Helm install should (whenever possible) rely on the functionality provided by Helm and only use our own logic where Helm has no native support. This is important to make sure HelmController's behavior stays as close to Helm's as possible. 

From above tenets we can see that `install` and `upgrade` have different default values (for backward compatibility resp. to provide Helm behavior by default). Therefore, I am inclined towards not providing a global `crds` attribute 
and to go with the action specific `crds` attributes instead.

Let's extend the policy enum like this:

`crds := Skip | Create | CreateReplace`

```
spec:
  install:
    # Default: "Create"
    crds: "Create"
    # Default false (we have to keep this for backward compatibility and might remove it in a future API version)
    skipCRDs: false
  upgrade:
    # Default: "Skip"
    crds: "CreateReplace"
```

And we can implement the following logic:

|action|`crds` value|Behaviour|
|---|---|---|
|`install`|`Skip`|Same as `install.skipCRDs = true`, pass skipCRDs flag to Helm|
|`install`|`Create`|Same as `install.skipCRDs = false`, let Helm create the CRDs for us|
|`install`|`CreateReplace`|Use our own logic, as Helm does not support this natively|
|`upgrade`|`Skip`|Do nothing at all|
|`upgrade`|`Create`|Use our own logic, as Helm does not support this natively|
|`upgrade`|`CreateReplace`|Use our own logic, as Helm does not support this natively|

What should happen if the the user specifies contradicting `crds` and `skipCRDs` on `install`?

```
spec:
  install:
    crds: "Create"
    skipCRDs: true
```

Shall we fail with an appropriate error message (my preference) or shall we defined a precedence rule?

What do you think?

hiddeco: Sounds like a solid plan.

> What should happen if the the user specifies contradicting `crds` and `skipCRDs` on install?
> Shall we fail with an appropriate error message (my preference) or shall we defined a precedence rule?

Due to deprecation giving `crds` precedence over `skipCRDs` would be an OK thing to do. But returning an error will save some users but annoy others, and usually saving a minority at the cost of annoying another is the right thing to do.

alex-berger: OK, I implemented this, ready for review.

docs/spec/v2beta1/helmreleases.md
```diff
@@ -1109,6 +1154,57 @@ spec:
...
+[offical Helm documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts).
+
+However, if you have to integrate and use many existing (upstream) Helm Charts, not being able to
+upgrade the CRDs via FluxCD `HelmRelease` objects might become a cumbersome limitation within your GitOps
+workflow. Therefore, FluxCD allows you to opt-in to upgrading CRDs by setting the `crds` policy on
+the `HelmRelease.spec.install` and `HelmRelease.spec.upgrade` objects.
+The following UpgradeCRDs policies are supported:
+
+- `Skip` Skip CRDs do neither install nor replace (update) any CRDs.
+- `Create` Only create new CRDs which doe not yet exist, neither update nor delete any existing CRDs.
```
hiddeco, line 1179:
```suggestion
- `Create` Only create new CRDs which do not yet exist, neither update nor delete any existing CRDs.
```

internal/runner/runner.go
```diff
@@ -131,14 +166,174 @@ func (r *Runner) Upgrade(hr v2.HelmRelease, chart *chart.Chart, values chartutil
...
+	return policy, nil
+}
+
+type rootScoped struct{}
+
+func (*rootScoped) Name() meta.RESTScopeName {
+	return meta.RESTScopeNameRoot
+}
+
+// This has been adapte from https://github.com/helm/helm/blob/v3.5.4/pkg/action/install.go#L127
```
hiddeco, line 213:
```suggestion
// This has been adapted from https://github.com/helm/helm/blob/v3.5.4/pkg/action/install.go#L127
```

hiddeco (commented):
Looks like the tests do now work?

alex-berger:
> Looks like the tests do now work?

As mentioned [above](https://github.com/fluxcd/helm-controller/pull/250#issuecomment-822422660) the test run on branches and tags, but unfortunately not on pull-requests. See my feature branch (the source branch of the PR at https://github.com/alex-berger/helm-controller/actions for test results).

I have not yet figured out whether there is a way to reference a GitHub pull-request as branch (I wasted 6 hours trying to find a way, but did not find one yet). As my time is limited I have given up and disabled the (new) tests for pull-requests, such that they only run on branches (and maybe tags).

hiddeco (approved):
I am OK with the e2e approach for now, we should move many of the tests to Go anyway, as has been done with a lot of kustomize-controller.

Anyway, thanks a lot @alex-berger. :100: :sunflower:

internal/runner/runner.go
```diff
@@ -58,7 +70,7 @@ type Runner struct {
 // namespace configured to the provided values.
 func NewRunner(getter genericclioptions.RESTClientGetter, storageNamespace string, logger logr.Logger) (*Runner, error) {
 	runner := &Runner{
-		logBuffer: NewLogBuffer(NewDebugLog(logger).Log, 0),
+		logBuffer: NewLogBuffer(NewDebugLog(logger).Log, 100),
```
hiddeco, line 73: I somehow missed this during review, and this change is not related to any of the CRD things. Can you explain why this was changed? As it changes the default of including the last 5 log lines (after de-duplication) in a failure condition and/or event to 100, which makes any failed event we produce now quite verbose.

alex-berger: I increased it for debugging during feature development and eventually forgot to decrease it again. 😳

hiddeco: Thanks, that means my restore PR is justified and I can continue with the release :-)

joejulian:
Looks like I'm moments late to comment on this, but has the justifications for helm, itself, not supporting this been considered? https://github.com/helm/community/blob/main/hips/hip-0011.md

hiddeco:
They have been considered, which is why we default to not enable the behavior as it can be potentially disastrous.

However, given that people use the helm-controller to automatically drive operations, having access to a configuration option that does the right think if you know how your chart (or the application behind it) behaves on Custom Resource Definition upgrades, it is in my opinion justified.

joejulian:
This feature would need to be able to add the conversion webhooks simultaneously. Links to them are defined in the CRD itself: see the [cert-manager CRDs](https://github.com/jetstack/cert-manager/blob/master/deploy/crds/crd-certificates.yaml#L26-L36) for example. Without them, the API server will reject create/change verbs to those CRDs.

PR #266, opened 2021-05-22
Fix HelmChartTemplateSpec Doc missing valuesFiles info

filintod:
The spec docs differs from the api docs, missing the information about ValuesFiles. This also is not shown in https://fluxcd.io/docs/components/helm/helmreleases/#specification that I guess comes from this file and why I realized it was no there. I'm not sure if there is an automation for this file or not but did not find it, I saw one for the api doc one.

Signed-off-by: Filinto Duran <duranto@gmail.com>


stefanprodan (approved):
LGTM 

Thanks @filintod

PR #271, opened 2021-06-01
Add support for Helm `--wait-for-jobs` flag

kewats:
Adding support for setting wait-for-jobs in HelmRelease.
 
https://github.com/helm/helm/issues/5948

### DIFF
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -319,6 +319,11 @@ type Install struct {
 	// +optional
 	DisableWait bool `json:"disableWait,omitempty"`
 
+	// DisableWaitForJobs disables waiting for jobs to complete after a Helm
+	// install has been performed.
+	// +optional
+	DisableWaitForJobs bool `json:"disableWaitForJobs,omitempty"`
+
 	// DisableHooks prevents hooks from running during the Helm install action.
 	// +optional
 	DisableHooks bool `json:"disableHooks,omitempty"`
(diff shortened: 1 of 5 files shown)


hiddeco (commented):
Thank you for your pull request :sunflower: 

Think it would be better to enable this by default (as we do with `--wait`), and make the flag an "inverse", so e.g. `DisableWaitForJobs`. This ensures people get the best automated experience by default.

kewats:
Sure, I will make the change. I agree that gives a better experience by default.

hiddeco:
When you are done with the changes, please smash your commits into one as the change is quite small. Thank you :coffee:

docs/spec/v2alpha1/helmreleases.md
```diff
@@ -142,6 +142,11 @@ type Install struct {
 	// +optional
```
hiddeco, line 142: This document is kept for historical reason, but should not include any changes as the `v2alpha1` has never supported these fields. Instead, they should be documented in `v2beta1`.

kewats:
Hi @hiddeco, I have inverted the flag and squashed my commits. Please review, thanks!

docs/spec/v2alpha1/helmreleases.md
kewats: Done!

hiddeco (approved):
LGTM, thanks a lot @kewats :1st_place_medal: :100: 

~I think your PR needs a tiny rebase with `main` so that I am able to merge it without any conflicts.~

PR #280, opened 2021-06-22
Fix HelmRelease Values merging strategy when there's a TargetPath Array reference

samueltorres:
When using the valuesFrom.targetPath property with an array index on HelmReleases, helm-controller is not merging correctly the content of the values <> valuesFrom properties.

This PR aims to fix the issue:
https://github.com/fluxcd/helm-controller/issues/281

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -521,7 +521,10 @@ func (r *HelmReleaseReconciler) getServiceAccountToken(ctx context.Context, hr v
 // and merges them as defined. Referenced resources are only retrieved once
 // to ensure a single version is taken into account during the merge.
 func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRelease) (chartutil.Values, error) {
-	result := chartutil.Values{}
+	result := hr.GetValues()
+	if result == nil {
+		result = chartutil.Values{}
+	}
 
 	configMaps := make(map[string]*corev1.ConfigMap)
 	secrets := make(map[string]*corev1.Secret)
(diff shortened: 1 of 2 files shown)


hiddeco:
This completely changes the behavior as documented, given the `spec.values` should be merged in last (overwriting any earlier values), where you are now using them as a base.

samueltorres:
Yes, you are totally right. My bad for not seeing it.
By the way, why was it designed like that ? :) 
Do you know any other way I can achieve what I've described in the issue? 
I am setting up some values and want to get some properties from some secrets, but this issue kind of throws that idea away. 
If I use the helm command setting the -set ... = ... it works as I wanted to.

hiddeco:
> By the way, why was it designed like that ? :)

Because `ConfigMap` and `Secret` references can theoretically be shared, and the idea is that if they are shared, any values there can always be overwritten with a declaration in the `HelmRelease`, which is the "truest" declaration of state.

> I am setting up some values and want to get some properties from some secrets, but this issue kind of throws that idea away.
> If I use the helm command setting the -set ... = ... it works as I wanted to.

There is little difference in how we merge values and how Helm `--set` works. The problem you are having is likely that when you use `helm`, you define some other values that build up part of the values structure, but now aren't present (because they are in the `spec.values`).

To work around this, you can try configure the base structure in a higher level `valuesFrom` item, [I expect this to work as our tests do the same](https://github.com/fluxcd/helm-controller/blob/main/config/testdata/valuesfrom/helmrelease.yaml#L33).

samuel-form3:
Thank you for the explanation, you helped me a lot. Although I feel that perhaps the merging package could support the thing I have in my issue.

samueltorres:
Thank you for the explanation, you helped me a lot. Although I feel that perhaps the merging package could support the thing I have in my issue.

PR #287, opened 2021-06-29
e2e: Update Kubernetes to v1.21

stefanprodan:


.github/workflows/e2e.yaml
```diff
@@ -561,3 +562,7 @@ jobs:
           helm ls -n helm-system --all
           kubectl -n helm-system logs deploy/source-controller
           kubectl -n helm-system logs deploy/helm-controller
+
```
squaremo, line 565: What's this bit?

stefanprodan: Goland trying to be smart :) removed

squaremo (approved):
LRTM

PR #298, opened 2021-07-17
Treat quoted values as string when targetPath is set

laozc:
When the values contain quoted strings, use ParseIntoString to enforce they are set as strings.
This allows a similar behavior with CLI --set-string.


hiddeco (commented):
Thank you for your contribution! :pineapple: 

Would you be able to add at least one test to https://github.com/fluxcd/helm-controller/blob/961856236f57fd41e9520c3e9ce8aad52611b0b7/controllers/helmrelease_controller_test.go#L37 to confirm that what you are adding here does work properly?

laozc:
> Thank you for your contribution! 🍍
> 
> Would you be able to add at least one test to
> 
> https://github.com/fluxcd/helm-controller/blob/961856236f57fd41e9520c3e9ce8aad52611b0b7/controllers/helmrelease_controller_test.go#L37
> 
> to confirm that what you are adding here does work properly?

Sure. Done.

hiddeco (approved):
Thank you @laozc :100:

PR #305, opened 2021-07-31
Remove old util ObjectKey

darkowlzz:
Replace old utility function ObjectKey with controller-runtime's
client.ObjectKeyFromObject.


hiddeco (approved):
Thanks Sunny 🌻

PR #306, opened 2021-08-04
Make the kubeconfig secrets compatible with SOPS

stefanprodan:
Add `values.yaml` to the supported kubeconfig secret key names in order for SOPS to correctly detect the storage format based on the file extension.

Ref: https://github.com/fluxcd/kustomize-controller/pull/400


hiddeco (approved):
Thank you @stefanprodan :pear:

PR #310, opened 2021-08-09
Revert "Merge pull request #294 from makkes/default-observed-generation"

hiddeco:
It is likely this new default trips up existing `HelmRelease` resources as the change seems to reset existing state.

If so, we need to figure out how to avoid this before reattempting.


hiddeco:
Report seems to be false, phew.

PR #342, opened 2021-10-19
Refactor controller with pkg runtime helpers

### LINKED ISSUE #322: Implement pkg/runtime helpers
As [part of the refactor work to set out some standards for the controllers that are part of the Flux family](https://github.com/fluxcd/flux2/issues/1601), the `pkg/runtime` helpers and interfaces should be implemented to work with the conditions, metrics and events recorded for HelmRelease resources.

In short, (at least) the following should be implemented:

- [ ] `HelmRelease` should implement the interfaces from [`conditions`](https://pkg.go.dev/github.com/fluxcd/pkg/runtime@v0.13.0-rc.5/conditions), and use this package to work with conditions.
- [ ] `HelmReleaseReconciler` should implement the helpers from [`controller`](https://pkg.go.dev/github.com/fluxcd/pkg/runtime@v0.13.0-rc.5/controller), and use these to record Events and Metrics.
- [ ] `HelmReleaseReconciler` should exit early without recording _any_ status when `.spec.suspend` is defined.
- [ ] The [`patch`](https://pkg.go.dev/github.com/fluxcd/pkg/runtime@v0.13.0-rc.5/patch) package should be implemented to (safely) patch resources without running into conflicts.
- [ ] The [`kstatus`](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) `Reconciling` and `Stalled` conditions should be properly implemented to provide better signals.

More examples can be found at:

- https://github.com/fluxcd/source-controller/pull/413
- https://github.com/fluxcd/image-automation-controller/pull/218

phillebaba:
Fixes #322 

- [x] HelmRelease should implement the interfaces from conditions, and use this package to work with conditions.
- [x] HelmReleaseReconciler should implement the helpers from controller, and use these to record Events and Metrics.
- [x] HelmReleaseReconciler should exit early without recording any status when .spec.suspend is defined.
- [x] The patch package should be implemented to (safely) patch resources without running into conflicts.
- [x] The kstatus Reconciling and Stalled conditions should be properly implemented to provide better signals.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -162,6 +162,7 @@ func (r *HelmReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			obj.Status.SetLastHandledReconcileRequest(v)
 		}
 
+		// TODO: Handle test condition when tests are enabled
 		// Summarize Ready condition
 		conditions.SetSummary(obj,
 			meta.ReadyCondition,
@@ -169,9 +170,13 @@ func (r *HelmReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				v2.ReleasedCondition,
 				//v2.TestSuccessCondition,
 				v2.RemediatedCondition,
+				v2.InitFailedCondition,
+				v2.RetriesExhaustedCondition,
 			),
 			conditions.WithNegativePolarityConditions(
 				v2.RemediatedCondition,
+				v2.InitFailedCondition,
+				v2.RetriesExhaustedCondition,
 			),
 		)
 
@@ -182,6 +187,8 @@ func (r *HelmReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 					v2.ReleasedCondition,
 					v2.TestSuccessCondition,
 					v2.RemediatedCondition,
+					v2.InitFailedCondition,
+					v2.RetriesExhaustedCondition,
 					meta.ReadyCondition,
 					meta.ReconcilingCondition,
 					meta.StalledCondition,
@@ -242,11 +249,14 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, obj *v2.HelmRelea
 	// Mark the resource as under reconciliation
 	conditions.MarkReconciling(obj, meta.ProgressingReason, "")
 
+	// Remove InitFailedCondition
+	conditions.Delete(obj, v2.InitFailedCondition)
+
 	// Reconcile chart based on the HelmChartTemplate
 	hc, reconcileErr := r.reconcileChart(ctx, obj)
 	if reconcileErr != nil {
 		obj.IncrementFailureCounter()
-		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "Chart reconcilliation failed: %s", reconcileErr.Error())
+		conditions.MarkTrue(obj, v2.InitFailedCondition, v2.ArtifactFailedReason, "Chart reconcilliation failed: %s", reconcileErr.Error())
 		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "Chart reconcilliation failed: %s", reconcileErr.Error())
 		return ctrl.Result{Requeue: true}, reconcileErr
 	}
@@ -254,7 +264,7 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, obj *v2.HelmRelea
 	if !conditions.IsReady(hc) {
 		msg := fmt.Sprintf("HelmChart '%s/%s' is not ready", hc.GetNamespace(), hc.GetName())
 		obj.IncrementFailureCounter()
-		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, msg)
+		conditions.MarkTrue(obj, v2.InitFailedCondition, v2.ArtifactFailedReason, msg)
 		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, msg)
 		log.Info(msg)
 		// Do not requeue immediately, when the artifact is created
@@ -267,8 +277,8 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, obj *v2.HelmRelea
 		if err := r.checkDependencies(ctx, obj); err != nil {
 			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s", err.Error(), r.requeueDependency.String())
 			obj.IncrementFailureCounter()
-			conditions.MarkFalse(obj, meta.ReadyCondition, v2.GetHelmChartFailedReason, msg)
-			r.Eventf(ctx, obj, events.EventSeverityError, v2.GetHelmChartFailedReason, msg)
+			conditions.MarkTrue(obj, v2.InitFailedCondition, v2.DependencyNotReadyReason, msg)
+			r.Eventf(ctx, obj, events.EventSeverityError, v2.DependencyNotReadyReason, msg)
 			log.Info(msg)
 			// Exponential backoff would cause execution to be prolonged too much,
 			// instead we requeue on a fixed interval.
@@ -281,7 +291,7 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, obj *v2.HelmRelea
 	values, err := r.composeValues(ctx, obj)
 	if err != nil {
 		obj.IncrementFailureCounter()
-		conditions.MarkFalse(obj, meta.ReadyCondition, v2.InitFailedReason, "could not get chart values: %s", err.Error())
+		conditions.MarkTrue(obj, v2.InitFailedCondition, v2.InitFailedReason, "could not get chart values: %s", err.Error())
 		r.Eventf(ctx, obj, events.EventSeverityError, v2.InitFailedReason, "could not get chart values: %s", err.Error())
 		return ctrl.Result{Requeue: true}, nil
 	}
@@ -290,16 +300,16 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, obj *v2.HelmRelea
 	chart, err := r.loadHelmChart(hc)
 	if err != nil {
 		obj.IncrementFailureCounter()
-		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
-		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
+		conditions.MarkTrue(obj, v2.InitFailedCondition, v2.ArtifactFailedReason, "could not load chart: %s", err.Error())
+		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "could not load chart: %s", err.Error())
 		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// Reconcile Helm release
-	if result, err := r.reconcileRelease(ctx, obj, chart, values); reconcileErr != nil {
+	if result, err := r.reconcileRelease(ctx, obj, chart, values); err != nil {
 		obj.IncrementFailureCounter()
-		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
-		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
+		// TODO: Set a better reason
+		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "reconciliation failed: %s", err.Error())
 		return result, err
 	}
 	return ctrl.Result{RequeueAfter: obj.Spec.Interval.Duration}, nil
@@ -312,14 +322,14 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, hr *v2.Hel
 	getter, err := r.getRESTClientGetter(ctx, hr)
 	if err != nil {
 		hr.IncrementFailureCounter()
-		conditions.MarkFalse(hr, meta.ReadyCondition, v2.InitFailedReason, err.Error())
+		conditions.MarkTrue(hr, v2.InitFailedCondition, v2.InitFailedReason, err.Error())
 		r.Eventf(ctx, hr, events.EventSeverityError, v2.InitFailedReason, err.Error())
 		return ctrl.Result{}, err
 	}
 	run, err := runner.NewRunner(getter, hr.GetStorageNamespace(), log)
 	if err != nil {
 		hr.IncrementFailureCounter()
-		conditions.MarkFalse(hr, meta.ReadyCondition, v2.InitFailedReason, "failed to initialize Helm action runner: %s", err)
+		conditions.MarkTrue(hr, v2.InitFailedCondition, v2.InitFailedReason, "failed to initialize Helm action runner: %s", err)
 		r.Eventf(ctx, hr, events.EventSeverityError, v2.InitFailedReason, "failed to initialize Helm action runner: %s", err)
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


controllers/helmrelease_controller_chart.go
```diff
@@ -174,10 +162,10 @@ func buildHelmChartFromTemplate(hr *v2.HelmRelease) *sourcev1.HelmChart {
 				Name: template.Spec.SourceRef.Name,
 				Kind: template.Spec.SourceRef.Kind,
 			},
-			Interval:          template.GetInterval(hr.Spec.Interval),
-			ReconcileStrategy: template.Spec.ReconcileStrategy,
-			ValuesFiles:       template.Spec.ValuesFiles,
-			ValuesFile:        template.Spec.ValuesFile,
+			Interval:    template.GetInterval(hr.Spec.Interval),
+			ValuesFiles: template.Spec.ValuesFiles,
+			ValuesFile:  template.Spec.ValuesFile,
```
phillebaba, line 167: Should Suspend and AccessFrom be added to here?

controllers/helmrelease_controller.go
```diff
@@ -128,271 +132,299 @@ func (c ConditionError) Error() string {
...
-		hr = v2.HelmReleaseProgressing(hr)
-		if updateStatusErr := r.patchStatus(ctx, &hr); updateStatusErr != nil {
-			log.Error(updateStatusErr, "unable to update status after state update")
-			return hr, updateStatusErr
-		}
-		// Record progressing status
-		r.recordReadiness(ctx, hr)
-	}
+	// TODO: Progressing is set immediately on reconcile so is it necessary to set it here?
+	hr, _ = v2.HelmReleaseAttempted(hr, revision, releaseRevision, valuesChecksum)
```
phillebaba, line 334: Should hasNewState be used here?

api/v2beta1/helmrelease_types.go
```diff
@@ -824,36 +814,9 @@ func (in HelmReleaseStatus) GetHelmChart() (string, string) {
...
-	meta.SetResourceCondition(&hr, meta.ReadyCondition, metav1.ConditionUnknown, meta.ProgressingReason,
-		"Reconciliation in progress")
-	resetFailureCounts(&hr)
-	return hr
-}
-
-// HelmReleaseNotReady registers a failed reconciliation of the given HelmRelease.
-func HelmReleaseNotReady(hr HelmRelease, reason, message string) HelmRelease {
-	meta.SetResourceCondition(&hr, meta.ReadyCondition, metav1.ConditionFalse, reason, message)
-	hr.Status.Failures++
```
phillebaba, line 841: Need to implement failure counter back into the logic.

api/v2beta1/helmrelease_types.go
```diff
@@ -106,7 +105,7 @@ type HelmReleaseSpec struct {
 	// references to HelmRelease resources that must be ready before this HelmRelease
 	// can be reconciled.
 	// +optional
-	DependsOn []dependency.CrossNamespaceDependencyReference `json:"dependsOn,omitempty"`
+	DependsOn []meta.NamespacedObjectReference `json:"dependsOn,omitempty"`
```
darkowlzz, line 108: The comment above this still mentions `dependency.CrossNamespaceDependencyReference`.

phillebaba: Fixed

controllers/helmrelease_controller.go
```diff
@@ -128,271 +132,299 @@ func (c ConditionError) Error() string {
...
-	if len(hr.Spec.DependsOn) > 0 {
-		if err := r.checkDependencies(hr); err != nil {
-			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s",
-				err.Error(), r.requeueDependency.String())
-			r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityInfo, msg)
+	if len(obj.Spec.DependsOn) > 0 {
+		if err := r.checkDependencies(ctx, obj); err != nil {
+			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s", err.Error(), r.requeueDependency.String())
+			// TODO: Where did meta.DependencyNotReadyReason go?
+			conditions.MarkFalse(obj, meta.ReadyCondition, meta.FailedReason, msg)
```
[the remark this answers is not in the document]
hiddeco: See the message of the commit which removed it: https://github.com/fluxcd/pkg/commit/4a7e889999fe22108af011e92ca766d0af5a48b3

controllers/helmrelease_controller.go
hiddeco: To keep most of the logic (that should be rewritten as well as noted in (an) other issue(s)) intact I would maintain it for now.

controllers/helmrelease_controller.go
```diff
@@ -128,271 +132,299 @@ func (c ConditionError) Error() string {
...
+			_, testErr := run.Test(*hr)
+			testErr = r.handleHelmActionResult(ctx, hr, revision, testErr, "test",
 				v2.TestSuccessCondition, v2.TestSucceededReason, v2.TestFailedReason)
 
 			// Propagate any test error if not marked ignored.
 			if testErr != nil && !remediation.MustIgnoreTestFailures(hr.Spec.GetTest().IgnoreFailures) {
-				testsPassing := apimeta.FindStatusCondition(hr.Status.Conditions, v2.TestSuccessCondition)
-				meta.SetResourceCondition(&hr, v2.ReleasedCondition, metav1.ConditionFalse, testsPassing.Reason, testsPassing.Message)
+				// TODO: Should this be checked for nil or just expected to be present
+				testsPassing := conditions.Get(hr, v2.TestSuccessCondition)
```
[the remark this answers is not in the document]
hiddeco: It should eventually not rely on a previously recorded condition at all, but to add a bit more safety checking right now, it wouldn't be wrong to add an explicit `nil` check.

controllers/helmrelease_controller_chart.go
phillebaba: I think the suspend feature should be in its own PR to keep things simple. But yes the question about setting both fields is an important one. Is it due to legacy that we have both field options or is there some other reason behind this? My suggestion for now is to just throw an error if both are set.

api/v2beta1/helmrelease_types.go
```diff
@@ -246,15 +245,6 @@ type HelmChartTemplateSpec struct {
...
 	Interval *metav1.Duration `json:"interval,omitempty"`
 
-	// Determines what enables the creation of a new artifact. Valid values are
-	// ('ChartVersion', 'Revision').
-	// See the documentation of the values for an explanation on their behavior.
-	// Defaults to ChartVersion when omitted.
-	// +kubebuilder:validation:Enum=ChartVersion;Revision
-	// +kubebuilder:default:=ChartVersion
-	// +optional
-	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
```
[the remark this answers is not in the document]
phillebaba: My guess is that the answer is yes.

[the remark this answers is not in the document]
hiddeco: Missed this comment for some reason :-S

No, it was introduced to the `HelmChart` resource in a later version than the `reconcilers-dev` branch. So you don't have access to the API against this version, but it will return.

controllers/helmrelease_controller_chart.go
stefanprodan: Given that the controller generates the HelmChart, `AccessFrom` should be set to allow the namespace where the HelmRelease is. Sadly it's impossible to to restrict the access to a single namespace on Kubernetes < 1.21 since the `kubernetes.io/metadata.name` label was aded in 1.21. Given this I would set this to allow all.

controllers/helmrelease_controller.go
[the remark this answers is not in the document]
phillebaba: Implemented a specific reason for this.

api/v2beta1/helmrelease_types.go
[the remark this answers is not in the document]
phillebaba: Ok good, will keep this open and loop back when it becomes relevant.

api/v2beta1/helmrelease_types.go
phillebaba: Failure counter should be implemented now.

controllers/helmrelease_controller.go
```diff
@@ -128,271 +132,299 @@ func (c ConditionError) Error() string {
...
+		conditions.SetSummary(obj,
+			meta.ReadyCondition,
+			conditions.WithConditions(
+				v2.ReleasedCondition,
+				v2.TestSuccessCondition,
+				v2.RemediatedCondition,
+			),
+			conditions.WithNegativePolarityConditions(
+				v2.RemediatedCondition,
+			),
```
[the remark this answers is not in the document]
phillebaba: So for now it looks right, but I will have to come back in a later PR and make changes to the different conditions.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
+		r.Eventf(ctx, obj, events.EventSeverityError, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// Reconcile Helm release
-	reconciledHr, reconcileErr := r.reconcileRelease(ctx, *hr.DeepCopy(), chart, values)
-	if reconcileErr != nil {
-		r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityError,
-			fmt.Sprintf("reconciliation failed: %s", reconcileErr.Error()))
+	if result, err := r.reconcileRelease(ctx, obj, chart, values); reconcileErr != nil {
```
darkowlzz, line 299: I've been investigating the reason for the test failure in CI and found that because of `reconcileErr` here, the error from `reconcileRelease()` is always ignored. 

```suggestion
	if result, err := r.reconcileRelease(ctx, obj, chart, values); err != nil {
```

darkowlzz: This doesn't resolve the CI failures, there's more to it. Investigating...

darkowlzz: This should fix most of the CI failures, but I found some more conditionals related issues. Will comment them inline.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
+
 	// Reconcile chart based on the HelmChartTemplate
-	hc, reconcileErr := r.reconcileChart(ctx, &hr)
+	hc, reconcileErr := r.reconcileChart(ctx, obj)
 	if reconcileErr != nil {
-		msg := fmt.Sprintf("chart reconciliation failed: %s", reconcileErr.Error())
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, msg)
-		return v2.HelmReleaseNotReady(hr, v2.ArtifactFailedReason, msg), ctrl.Result{Requeue: true}, reconcileErr
+		obj.IncrementFailureCounter()
+		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "Chart reconcilliation failed: %s", reconcileErr.Error())
```
darkowlzz, line 249: Since ready condition is calculated as a summary based on other conditions, it'd be better to mark `ReleasedCondition` false here. Setting ready condition here gets overwritten in the `Reconcile()` when ready condition is set as a summary of other conditions.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
+		return ctrl.Result{Requeue: true}, reconcileErr
 	}
 
-	// Check chart readiness
-	if hc.Generation != hc.Status.ObservedGeneration || !apimeta.IsStatusConditionTrue(hc.Status.Conditions, meta.ReadyCondition) {
+	if !conditions.IsReady(hc) {
 		msg := fmt.Sprintf("HelmChart '%s/%s' is not ready", hc.GetNamespace(), hc.GetName())
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityInfo, msg)
+		obj.IncrementFailureCounter()
+		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, msg)
```
darkowlzz, line 257: Same as above, this can also be `ReleasedCondition` marked as false.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
-	if len(hr.Spec.DependsOn) > 0 {
-		if err := r.checkDependencies(hr); err != nil {
-			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s",
-				err.Error(), r.requeueDependency.String())
-			r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityInfo, msg)
+	if len(obj.Spec.DependsOn) > 0 {
+		if err := r.checkDependencies(ctx, obj); err != nil {
+			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s", err.Error(), r.requeueDependency.String())
+			obj.IncrementFailureCounter()
+			conditions.MarkFalse(obj, meta.ReadyCondition, v2.GetHelmChartFailedReason, msg)
```
darkowlzz, line 270: Same as above, this can also be `ReleasedCondition` marked as false.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
 	}
 
 	// Compose values
-	values, err := r.composeValues(ctx, hr)
+	values, err := r.composeValues(ctx, obj)
 	if err != nil {
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, err.Error())
-		return v2.HelmReleaseNotReady(hr, v2.InitFailedReason, err.Error()), ctrl.Result{Requeue: true}, nil
+		obj.IncrementFailureCounter()
+		conditions.MarkFalse(obj, meta.ReadyCondition, v2.InitFailedReason, "could not get chart values: %s", err.Error())
```
darkowlzz, line 284: Same as above, this can also be `ReleasedCondition` marked as false.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// Load chart from artifact
 	chart, err := r.loadHelmChart(hc)
 	if err != nil {
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, err.Error())
-		return v2.HelmReleaseNotReady(hr, v2.ArtifactFailedReason, err.Error()), ctrl.Result{Requeue: true}, nil
+		obj.IncrementFailureCounter()
+		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
```
darkowlzz, line 293: Same as above, this can also be `ReleasedCondition` marked as false.

controllers/helmrelease_controller.go
```diff
@@ -128,170 +132,206 @@ func (c ConditionError) Error() string {
...
 	}
 
 	// Reconcile Helm release
-	reconciledHr, reconcileErr := r.reconcileRelease(ctx, *hr.DeepCopy(), chart, values)
-	if reconcileErr != nil {
-		r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityError,
-			fmt.Sprintf("reconciliation failed: %s", reconcileErr.Error()))
+	if result, err := r.reconcileRelease(ctx, obj, chart, values); reconcileErr != nil {
+		obj.IncrementFailureCounter()
+		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, "could not get chart artifact: %s", err.Error())
```
darkowlzz, line 301: Same as above, this can also be `ReleasedCondition` marked as false.

controllers/helmrelease_controller.go
```diff
@@ -300,99 +340,102 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
...
+				return ctrl.Result{}, err
 			}
 		}
 
 		// Fail if install retries are exhausted.
 		if hr.Spec.GetInstall().GetRemediation().RetriesExhausted(hr) {
+			hr.IncrementFailureCounter()
 			err = fmt.Errorf("install retries exhausted")
-			return v2.HelmReleaseNotReady(hr, released.Reason, err.Error()), err
+			conditions.MarkFalse(hr, meta.ReadyCondition, released.Reason, err.Error())
```
darkowlzz, line 366: This can be `RemediatedCondition`.

darkowlzz (commented):
I tried reproducing the e2e tests that are failing in the CI and left comments about what I did to make it work. Also left a few comments around the status conditions usage. Hope those comments will make it clear some of the conditionals usage patterns.

Issue #345, opened 2021-10-24
High Memory Usage after helm-controller v0.12.0 upgrade


Legion2:
I updated to helm-controller v0.12.1 and started using `ReconcileStrategy` `Revision` for all my local helm charts. Now helm-controller is restarted each time I push a commit to the GitRepository source, because helm-controller uses too much memory and is killed by kubernetes (OOMKilled). As a result of the controller being killed by kubernetes, some helm release are stuck in the upgrade process which must be manually rolled back (https://github.com/helm/helm/issues/8987).
![image](https://user-images.githubusercontent.com/12752323/138603834-201d4897-cb9e-474a-bf32-5fdc3010781d.png)

tehlers320:
I'm not sure this is just the revision strategy. We do not use this anywhere and have seen a huge jump in memory use.  I grabbed a pprof heap map. 
![out](https://user-images.githubusercontent.com/16598119/139885982-3255e200-c952-4db5-aa8e-8e3d74052339.png)

hiddeco:
@tehlers320 can you provide more information about the version you upgraded from?

tehlers320:
@hiddeco we jumped from 0.10.2 and do not use the new reconcile revision.  Its been running for 5 days (when it can, OOMkiller). Sorry i don't mean to walk over this issue, but i think perhaps its not related to the reconcilestrategy since we dont use Revision. 

Should i make a new one or do you think the issue is both?

hiddeco:
No, I think your observations are correct based on other reports on Slack.

Did a quick dive into it with the limited time I had available, but the helm-controller didn't really change much besides Helm, kustomize and controller-runtime updates. It would be useful if someone could pinpoint the resource behavior change to an exact helm-controller version, which would help identifying the issue.

I am at present working on Helm improvement changes for the source-controller in the area of Helm repository index, dependency, and chart build memory consumption. Once that's done, I have time (and am planning) to look in much greater detail at the current shape of the helm-controller (as part of https://github.com/fluxcd/helm-controller/milestone/1).

glen-uc:
We are facing the same issue where the memory of helm controller keeps on growing

We upgraded from 0.16.2 (no issue with helm controller in this version) > 0.19.1 (started seeing issues ) > 0.20.1

We have around 258 helm releases and 8 helm repo's in our cluster with the interval set to 24h 

We have set 4GB as the memory limit for our helm controller

[...]

PR #352, opened 2021-11-03
Downgrade Helm to v3.6.3 due to OOM issues

stefanprodan:
Ref: https://github.com/fluxcd/helm-controller/issues/345#issuecomment-959104091


hiddeco (approved):
Based on user feedback of hard to replicate real world usage, this seems to be the one working for all.

The downside of this all is that we're stuck with Helm 3.6.3, but as Helm 3.7.0 didn't seem to have brought that many important features and/or fixes for us (mainly focused on OCI, which we do not have API access to anyway)

This seems to be the best solution to bring the stability back, buying us some time to either wait for upstream to settle things, or come up with a better solution.

PR #354, opened 2021-11-04
Replace kube-openapi with go-openapi

stefanprodan:
Alternative to #352


hiddeco:
This didn't appear to work and #352 is deemed the successor.

PR #358, opened 2021-11-12
Verify artifacts integrity

stefanprodan:
After downloading an artifact, compute its checksum and verify that it matches the original checksum as advertised by source-controller.

For backwards compatibility with source-controller v0.17.2, the checksum is computed with SHA1, on clusters with a newer source-controller we use SHA256.

Addresses https://github.com/fluxcd/source-controller/issues/468

### DIFF
--- a/controllers/helmrelease_controller_chart.go
+++ b/controllers/helmrelease_controller_chart.go
@@ -18,9 +18,10 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha1"
+	"crypto/sha256"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -94,7 +95,7 @@ func (r *HelmReleaseReconciler) getHelmChart(ctx context.Context, hr *v2.HelmRel
 // loads it into a chart.Chart, and removes the downloaded artifact.
 // It returns the loaded chart.Chart on success, or an error.
 func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*chart.Chart, error) {
-	f, err := ioutil.TempFile("", fmt.Sprintf("%s-%s-*.tgz", source.GetNamespace(), source.GetName()))
+	f, err := os.CreateTemp("", fmt.Sprintf("%s-%s-*.tgz", source.GetNamespace(), source.GetName()))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


controllers/helmrelease_controller_chart.go
```diff
@@ -126,13 +129,43 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to download artifact, error: %w", err)
+	}
+
+	// verify checksum matches origin
+	if err := r.verifyArtifact(source.GetArtifact(), bytes.NewBuffer(bodyBytes)); err != nil {
+		return nil, err
+	}
+
+	if _, err = io.Copy(f, bytes.NewBuffer(bodyBytes)); err != nil {
```
hiddeco, line 142: If you would use a `MultiWriter` in the `io.Copy` with a hasher, you wouldn't have to read all bytes, and wouldn't have to copy twice.

controllers/helmrelease_controller_chart.go
```diff
@@ -126,13 +128,42 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
...
+	}
+
+	if _, err = io.Copy(f, &buf); err != nil {
 		return nil, err
 	}
 
 	return loader.Load(f.Name())
 }
 
+func (r *HelmReleaseReconciler) verifyArtifact(artifact *sourcev1.Artifact, buf *bytes.Buffer, reader io.Reader) error {
```
hiddeco, line 145:
```suggestion
func (r *HelmReleaseReconciler) verifyArtifact(artifact *sourcev1.Artifact, writer io.Writer, reader io.Reader) error {
```

controllers/helmrelease_controller_chart.go
```diff
@@ -126,13 +128,42 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
...
+	var buf bytes.Buffer
+
+	// verify checksum matches origin
+	if err := r.verifyArtifact(source.GetArtifact(), &buf, resp.Body); err != nil {
+		return nil, err
+	}
+
+	if _, err = io.Copy(f, &buf); err != nil {
 		return nil, err
 	}
```
hiddeco, line 140:
```suggestion
```

controllers/helmrelease_controller_chart.go
```diff
@@ -126,13 +128,42 @@ func (r *HelmReleaseReconciler) loadHelmChart(source *sourcev1.HelmChart) (*char
 		return nil, fmt.Errorf("artifact '%s' download failed (status code: %s)", source.GetArtifact().URL, resp.Status)
 	}
 
-	if _, err = io.Copy(f, resp.Body); err != nil {
+	var buf bytes.Buffer
+
+	// verify checksum matches origin
+	if err := r.verifyArtifact(source.GetArtifact(), &buf, resp.Body); err != nil {
+		return nil, err
```
hiddeco, line 135:
```suggestion
	if err := r.verifyArtifact(source.GetArtifact(), f, resp.Body); err != nil {
		return nil, err
```

hiddeco (approved):
LGTM :1st_place_medal: :bow:

PR #360, opened 2021-11-16
Update Alpine to v3.14

vespian:
This mitigates alerts from security scanners, when scanning helm-controller docker image:
* http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3711
* http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-36159

Adding patchlevel version allows for more reproducible builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY internal/ internal/
 ENV CGO_ENABLED=0
 RUN xx-go build -a -o helm-controller main.go
 
-FROM alpine:3.13
+FROM alpine:3.14
 
 # link repo to the GitHub Container Registry image
 LABEL org.opencontainers.image.source="https://github.com/fluxcd/helm-controller"


stefanprodan:
> Adding patchlevel version allows for more reproducible builds.

It also means new CVE fixes will not land in Flux release unless someone bumps the patch version of Alpine in all our controllers. I’m not for doing this in Flux.

makkes:
Omitting the patch version worked pretty well in the past so I suppose let's leave it like that and just bump to 3.14.

stefanprodan:
@makkes we need to update Alpine to 3.14 for all the Flux components that use it as base image:
- kustomize-controller
- helm-controller
- notification-controller
-  image-reflector-controller
- flux-cli (Dockerfile in flux2 repo)

We use Debian in source-controller and image-automation-controller and that has many OS CVEs too.

vespian:
We may also consider closing this PR.

Alpine `3.13` is going to be supported until `2022-11-01`. If the policy is to use the latest Alpine image during the release time (i.e. do not specify patchlevel version in the container), then all recent flux releases that fetched `3.13.7` version are OK.

Bumping `3.14` makes sense if all flux containers should use the same Alpine minor release.

Let me know what you think.

stefanprodan:
I'm for bumping Alpine to `3.14` in all Flux controllers.

vespian:
> I'm for bumping Alpine to 3.14 in all Flux controllers and flux-cli.

ACK I adjusted the PR as requested. Please have another look.

makkes (approved):

makkes:
@vespian would you be up for bumping the Alpine version in the other controllers, too?

vespian:
> @vespian would you be up for bumping the Alpine version in the other controllers, too?

Sure thing.

vespian:
> @makkes we need to update Alpine to 3.14 for all the Flux components that use it as base image:

* kustomize-controller - already uses 3.14
* notification-controller - https://github.com/fluxcd/notification-controller/pull/285
* image-reflector-controller - https://github.com/fluxcd/image-reflector-controller/pull/203
* flux-cli - https://github.com/fluxcd/flux2/pull/2091

PR #361, opened 2021-11-19
Update various dependencies to mitigate CVE warnings

hiddeco:
Because even if there is no impact whatsoever, we are nice people.


stefanprodan (approved):
😢

PR #377, opened 2021-12-09
Update source-controller to `v0.19.2`

darkowlzz:


hiddeco (approved):
Do not think it's going to win the race from the source-controller CI, but once it turns green, this should be good to go :smile:

PR #380, opened 2021-12-10
Update Helm to v3.7.2

hiddeco:
This commit updates Helm to 3.7.2, in an attempt to get to a v3.7.x
release range _without_ any memory issues (see #345), which should have
been addressed in this release.

The change in replacements has been cross-checked with the dependencies
of Helm (and more specifically, the Oras project), and confirmed to not
trigger any warnings using `trivy`.

**:framed_picture: RC image:** `docker.io/fluxcd/helm-controller:rc-helm-3.7.2--a12eae8e`


darkowlzz (approved):

stefanprodan:
For those that prefer GHCR the image is at `ghcr.io/fluxcd/helm-controller:rc-helm-3.7.2--a12eae8e`

relu (approved):

stefanprodan:
We are going ahead with the Helm upgrade to v3.7.2 as v3.6.3 blocks us from fixing the containerd CVEs due to ORAS breaking changes. In case v3.7.2 brings back the memory leak reported in #345 please pin helm-controller to `v0.12.2`.

PR #385, opened 2021-12-21
security: Drop capabilities, set userid and enable seccomp

pjbgf:
Further restricts the `SecurityContext` that the controller runs under, by enabling the default seccomp profile, dropping all linux capabilities. 

This was set at container-level to ensure backwards compatibility with use cases in which more privileged sidecars are injected into the `source-controller` pod without setting less restrictive settings.

Note that seccomp will only be enabled if the container runtime and operational system supports it, otherwise the container will run `unconfined` (aka fail-open), which is the same behaviour as not setting the `seccompProfile` in the first place.

Relates to https://github.com/fluxcd/flux2/issues/2014
Co authored by @aryan9600

Impacts https://github.com/weaveworks/flux2-openshift/issues/10

BREAKING CHANGE: 
- The use of new seccomp API requires Kubernetes 1.19.
- The controller container is now executed under `65534:65534` (userid:groupid). This change may break deployments that hard-coded the user name 'controller' in their `PodSecurityPolicy`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,8 +41,6 @@ RUN apk add --no-cache ca-certificates tini
 
 COPY --from=builder /workspace/helm-controller /usr/local/bin/
 
-RUN addgroup -S controller -g 65532 && adduser -D -u 65532 -s /sbin/nologin -S controller -G controller
-
-USER controller
+USER 65534:65534
 
 ENTRYPOINT [ "/sbin/tini", "--", "helm-controller" ]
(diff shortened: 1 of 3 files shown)


pjbgf:
This was tested on Kubernetes `1.23.1` with restrict "pod security" (below) and it worked as expected. 

```
apiVersion: v1
kind: Namespace
metadata:
  name: flux-system
  labels:
    pod-security.kubernetes.io/enforce: restricted
    pod-security.kubernetes.io/enforce-version: latest
    pod-security.kubernetes.io/warn: restricted
    pod-security.kubernetes.io/warn-version: latest
```

stefanprodan (approved):
LGTM

PR #386, opened 2021-12-21
Fix inconsistent code-style raised at security audit

pjbgf:
Changes summary:
- Add check for nil to fix inconsistent code-style raised at the security audit.
- Short-circuit function when `ExternalEventRecorder` is nil.

Relates to https://github.com/fluxcd/pkg/issues/173


stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #391, opened 2022-01-11
Use go install instead of go get in Makefile

aryan9600:
Signed-off-by: Sanskar Jaiswal <sanskar.jaiswal@weave.works>


stefanprodan (approved):
LGTM

Thanks @aryan9600

PR #392, opened 2022-01-11
Use go-install-tool for gen-crd-api-reference-docs

aryan9600:
Signed-off-by: Sanskar Jaiswal <sanskar.jaiswal@weave.works>

Ref: https://github.com/fluxcd/flux2/issues/2273


hiddeco (approved):
LGTM. Thank you @aryan9600 :bow:

PR #395, opened 2022-01-14
Use patch instead of update when adding finalizers

kingdonb:
Following example of fluxcd/source-controller#542


stefanprodan (approved):
LGTM 

Thanks @kingdonb

PR #406, opened 2022-01-27
Allow setting a default service account for impersonation

stefanprodan:
Introduce the flag `--default-service-account` for allowing cluster admins to enforce impersonation across the cluster without having to use an admission controller. 

When the flag is set to a value that's not empty string, all HelmReleases which don't have `spec.serviceAccountName` specified will use the service account name provided by `--default-service-account=<SA Name>` in the namespace of the object.

**Breaking changes**: 
- When `--default-service-account=<SA name>` and/or `spec.ServiceAccountName` are specified, the controller no longer queries the API to extract and use the SA token. Instead, the controller relies on the Kubernetes impersonation feature to run the reconciliation under the `system:serviceaccount:<SA Namespace>:<SA Name>` identity. This impacts only clusters with custom made tokens and `automountServiceAccountToken` disabled. If a service account token points to a different identity (which should never be the case), then the controller will fail to reconcile with `Forbidden : User "system:serviceaccount:my-namespace:my-sa-name" cannot create resource`.
- When both `spec.kubeConfig` and `spec.ServiceAccountName` are specified, the controller will impersonate the service account on the target cluster, previously the controller ignored the service account. 

Part of: https://github.com/fluxcd/flux2/issues/2340

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -32,31 +32,48 @@ func NewInClusterRESTClientGetter(cfg *rest.Config, namespace string) genericcli
 	flags.BearerToken = &cfg.BearerToken
 	flags.CAFile = &cfg.CAFile
 	flags.Namespace = &namespace
+	if sa := cfg.Impersonate.UserName; sa != "" {
+		flags.Impersonate = &sa
+	}
+
 	return flags
 }
 
 // MemoryRESTClientGetter is an implementation of the genericclioptions.RESTClientGetter,
 // capable of working with an in-memory kubeconfig file.
 type MemoryRESTClientGetter struct {
-	kubeConfig []byte
-	namespace  string
+	kubeConfig         []byte
+	namespace          string
+	impersonateAccount string
 }
 
-func NewMemoryRESTClientGetter(kubeConfig []byte, namespace string) genericclioptions.RESTClientGetter {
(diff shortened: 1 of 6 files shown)


internal/kube/client.go
```diff
@@ -40,17 +44,26 @@ func NewInClusterRESTClientGetter(cfg *rest.Config, namespace string) genericcli
...
 	namespace  string
+	cfg        *rest.Config
 }
 
-func NewMemoryRESTClientGetter(kubeConfig []byte, namespace string) genericclioptions.RESTClientGetter {
+func NewMemoryRESTClientGetter(kubeConfig []byte, namespace string, cfg *rest.Config) genericclioptions.RESTClientGetter {
 	return &MemoryRESTClientGetter{
 		kubeConfig: kubeConfig,
 		namespace:  namespace,
+		cfg:        cfg,
```
hiddeco, line 54: Instead of embedding the `cfg` here, I think it would be better to inject the (string) values we want to (optionally) overwrite. This would create a better boundary and prevent from e.g. accidentally returning `c.cfg` instead of `cfg` (and thereby leaking a cluster-wide config).

stefanprodan: I've replaced it with `impersonateAccount`.

hiddeco (approved):
Thanks for following through on requested changes :bow:

Issue #351, opened 2021-11-03
One failing HelmRelease seems to block all other releases from being installed/upgraded


ilya-git:
1. Have a HelmRelease "A" that is successfully installed in k8s via flux
2. Add one new HelmRelease  "B" with an error (I had a deployment with an image tag that does not exist)
3. Flux tries to create a new HelmRelease "B" which it never succeeds as pod is stuck in container creating (since image does not exist). HelmRelease "B" has an event: `reconciliation failed: upgrade retries exhausted`
4. Now if we do changes to HelmRelease "A" (e.g. a new image discovered by an image policy), it would be never upgraded, and this event appears: `Helm upgrade failed: another operation (install/upgrade/rollback) is in progress`

If I understand it correctly a HelmRelease "B" is now effectively blocking any HelmRelease install/update since it can never finishes.

I don't know whether it is a bug or not, but it would be nice to not have one failing helm release blocking the others.

hiddeco:
Can you try downgrading to `v0.11.2` to see if the same behavior happens?

We think the Helm `v3.7.x` release range contains some troublesome code originating from the Kubernetes core (through `kubectl`), causing memory (#345 #349) and locking issues due to the usage of i.a. various globals.

nickperkins:
I think this bug might be in helm. The operation stalls for whatever reason and remains in a pending state. Manual requests for helm to do an install/upgrade/rollback will fail as well.

The workaround I have for the moment is to delete the helm release secret that is stuck in pending. This then allows you to reattempt the operation.

ilya-git:
I have now updated to the latest flux that should use the latest helm where these problems seems to be resolved. We will run for some time and see if the issue arises again

ilya-git:
The same bug already happened, so flux 0.25.2 did not seem to fix the issue unfortunately...

artem-nefedov:
I'm seeing same problem on flux v0.25.3/helm-controller v0.15.0.

gladiatr72:
@ilya-git https://github.com/fluxcd/helm-controller/issues/149 is worth keeping an eye on.

stefanprodan:
This may be related to https://github.com/helm/helm/pull/10486

PR #407, opened 2022-01-27
Update helm to 3.7.2

cbuto:
After upgrading to Flux 0.25.3, we noticed the memory discussed in a few different issues fluxcd/helm-controller#349, fluxcd/helm-controller#351, and fluxcd/helm-controller#345 started happening again (after pinning the helm-controller to 0.12.2, the issue went away).

The original update to helm 3.7.2 was in this [PR](https://github.com/fluxcd/helm-controller/pull/380) but ended up getting downgraded to 3.7.1 in a subsequent PR.

Hopefully this update resolves the memory issues! 🎉 

Signed-off-by: Casey Buto <cbuto@d2iq.com>


makkes (approved):

stefanprodan:
@cbuto have you tested this change and the memory issue went away?

cbuto:
> @cbuto have you tested this change and the memory issue went away?

@stefanprodan I didn't get a chance to test it fully, going to give it a go now though and see what happens!

cbuto:
@stefanprodan unfortunately, it looks like a memory issue might still exist (at least in our tests) with the helm-controller configured with 1Gi and 1.5Gi of memory.

stefanprodan:
@cbuto could you please test with helm 3.8.0?

cbuto:
@stefanprodan I can give that a go for sure, I did see there is actually some more work being done addressing memory issues here https://github.com/helm/helm/pull/10486 that could be related (hard to tell right now)

hiddeco:
I suspect the PR you are referring to is indeed the thing that (hopefully) resolves this all, and also explains some of the other weird behavior that has been reported recently (about all releases failing).

PR #408, opened 2022-01-28
Allow disabling cross-namespace references

SomtochiAma:
Introduce the flag `--no-cross-namespace-refs` (defaults to false) for allowing cluster admins to disable cross-namespace references to sources.

When the controller is run with `--no-cross-namespace-refs=true` and `chart.spec.sourceRef` refers to a `GitRepository`, `Bucket` or `HelmRepository` in a different namespace than the `HelmRelease` object, the reconciliation will fail with the `AccessDenied` reason.

On access denied errors the controller logs and sends an event e.g.:

```
can't access 'GitRepository/my-other-namespace/my-source', cross-namespace references have been blocked
```

And the `HelmRelease` status `Ready` condition it set to:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-01-26T07:26:48Z"
    message: "can't access 'GitRepository/my-other-namespace/my-source', cross-namespace references have been blocked"
    reason: AccessDenied
    status: "False"
    type: Ready
```

Part of: https://github.com/fluxcd/flux2/issues/2337

### DIFF
--- a/controllers/helmrelease_controller_chart.go
+++ b/controllers/helmrelease_controller_chart.go
@@ -28,6 +28,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/fluxcd/pkg/runtime/acl"
 	"github.com/hashicorp/go-retryablehttp"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
@@ -45,6 +46,14 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
 	chartName := types.NamespacedName{
 		Namespace: hr.Spec.Chart.GetNamespace(hr.Namespace),
 		Name:      hr.GetHelmChartName(),
+	}
+
+	if r.NoCrossNamespaceRef && chartName.Namespace != hr.Namespace {
+		return nil, acl.AccessDeniedError(fmt.Sprintf("can't access '%s/%s', cross-namespace references have been blocked",
+			hr.Spec.Chart.Spec.SourceRef.Kind, types.NamespacedName{
+				Namespace: hr.Spec.Chart.Spec.SourceRef.Namespace,
+				Name:      hr.Spec.Chart.Spec.SourceRef.Name,
+			}))
 	}
 
 	// Garbage collect the previous HelmChart if the namespace named changed.
--- a/docs/spec/v2beta1/helmreleases.md
+++ b/docs/spec/v2beta1/helmreleases.md
@@ -687,6 +687,10 @@ Supported source types:
 The `HelmChart` is created in the same namespace as the `sourceRef`,
 with a name matching the `HelmRelease` `<metadata.namespace>-<metadata.name>`.
 
+> **Note** that on multi-tenant clusters,  platform admins can disable cross-namespace references
+> with the `--no-cross-namespace-refs=true` flag. When this flag is set, the helmrelease can only
+> refer to sources in the same  namespace as the helmrelease object.
+
 The `chart.spec.chart` can either contain:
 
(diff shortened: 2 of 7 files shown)


controllers/helmrelease_controller_chart.go
```diff
@@ -47,6 +48,11 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
 		Name:      hr.GetHelmChartName(),
 	}
 
+	if r.NoCrossNamespaceRef && chartName.Namespace != hr.Namespace {
```
stefanprodan, line 51: I don't think this is the right place for this, the chart shouldn't be created if the source is in a different namespace. We need to check the `sourceRef.namespace` way before it reaches this code.

SomtochiAma: The chart is created after below line.
https://github.com/fluxcd/helm-controller/blob/main/controllers/helmrelease_controller_chart.go#L66

stefanprodan: Ok, we could tell users about the thing they've set in the spec which is the source (HelmRepo/GitRepo/Bucket). Instead of the `can't access 'HelmChart/chart-ns/chart-name'` could be `HelmRepository/repo-ns/repo-name`.

docs/spec/v2beta1/helmreleases.md
```diff
@@ -678,6 +678,15 @@ The `spec.chart.spec.sourceRef` is a reference to an object managed by
 [revision](https://github.com/fluxcd/source-controller/blob/main/docs/spec/v1beta1/common.md#source-status)
 changes, it generates a Kubernetes event that triggers a new release.
 
+> **Note** that on multi-tenant clusters,  platform admins can disable cross-namespace references
+> with the `--no-cross-namespace-refs=true` flag. When this flag is set, the helmrelease can only
+> refer to sources in the same  namespace as the helmrelease object.
+
+> **Note** that on multi-tenant clusters, platform admins can disable cross-namespace references
```
stefanprodan, line 685: Please delete this.

Makefile
```diff
@@ -32,7 +32,7 @@ manager: generate fmt vet
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
-	go run ./main.go
+	go run ./main.go --no-cross-namespace-refs
```
stefanprodan, line 35: Please undo this change.

docs/spec/v2beta1/helmreleases.md
```diff
@@ -678,6 +678,10 @@ The `spec.chart.spec.sourceRef` is a reference to an object managed by
 [revision](https://github.com/fluxcd/source-controller/blob/main/docs/spec/v1beta1/common.md#source-status)
 changes, it generates a Kubernetes event that triggers a new release.
 
+> **Note** that on multi-tenant clusters,  platform admins can disable cross-namespace references
```
stefanprodan, line 681: Please move this below after "The `HelmChart` is created in the same namespace as the `sourceRef`".

controllers/helmrelease_controller_chart.go
```diff
@@ -47,6 +48,14 @@ func (r *HelmReleaseReconciler) reconcileChart(ctx context.Context, hr *v2.HelmR
...
+		return nil, acl.AccessDeniedError(fmt.Sprintf("can't access '%s/%s', cross-namespace references have beem blocked",
```
stefanprodan, line 52:
```suggestion
		return nil, acl.AccessDeniedError(fmt.Sprintf("can't access '%s/%s', cross-namespace references have been blocked",
```

stefanprodan (approved):
LGTM 

Thanks @SomtochiAma

PR #409, opened 2022-01-28
Update Helm to patched 3.8.0

### LINKED ISSUE #351: One failing HelmRelease seems to block all other releases from being installed/upgraded

hiddeco:
This updates Helm to a manually patched 3.8.0 release to include the
memory/goroutine leak fix: https://github.com/hiddeco/helm/commits/3.8.x

Potentially fixes #351 

**Multi-arch test image:** `docker.io/hiddeco/helm-controller:update-helm-a919ad5`


cbuto:
✅ I tested this image under the same circumstances where we saw the helm-controller get OOMKilled before and the helm-controller was **not** OOMKilled (everything worked as expected 🎉 ). Since it was not OOMKilled, none of the HelmReleases got stuck with `Helm upgrade failed: another operation (install/upgrade/rollback) is in progress`.

go.mod
```diff
@@ -31,36 +30,36 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+replace helm.sh/helm/v3 v3.8.0 => github.com/hiddeco/helm/v3 v3.0.0-20220128105410-34ef0a7a5811
```
stefanprodan, line 33: Let's link to https://github.com/helm/helm/pull/10486 and explain that this replace is temporary until 3.8.1 is out.

stefanprodan (approved):
LGTM 🚀

PR #416, opened 2022-02-13
Add uninstall disableWait flag

samueltorres:
Add a uninstall `disableWait` property on Helm Release Uninstall API, so users can set whether Helm Controller waits for all the  resources to be deleted before finalizing the Helm Release.

This is very handy to ensure Helm Releases are fully un-installed. See [Slack Thread](https://cloud-native.slack.com/archives/CLAJ40HV3/p1644503412939859).

### DIFF
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -754,6 +754,11 @@ type Uninstall struct {
 	// release as deleted, but retain the release history.
 	// +optional
 	KeepHistory bool `json:"keepHistory,omitempty"`
+
+	// DisableWait disables waiting for all the resources to be deleted after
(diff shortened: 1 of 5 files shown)


hiddeco (commented):
Thank you for this PR, looks like it is a newly added flag introduced a couple Helm releases ago which I missed. :bow: 

I think the default behavior should be to wait (because that results in the best observation of an uninstall), to be inline with the other wait related flags for e.g. `Install` and `Upgrade`. The flag in the Custom Resource would then become `DisableWait`.

samueltorres:
Makes total sense :) Thanks for remind me that.

stefanprodan (approved):
LGTM

Thanks @samueltorres and @hiddeco

PR #421, opened 2022-02-21
Update `pkg/runtime` and `apis/meta`

darkowlzz:
This includes an update of the source-controller to v0.22.0, to pull in
the v1beta2 API which makes use of the same packages.


stefanprodan (commented):
You have to deploy SC v1beta2 for e2e testing, here is an example from KC https://github.com/fluxcd/kustomize-controller/blob/3117a4e8d20b80039e34ad55f0241cc2f866e060/config/default/kustomization.yaml

hiddeco (approved):

PR #422, opened 2022-02-21
Set QPS and Burst when impersonating service account

stefanprodan:
We don't map the client QPS and Burst setting resulting in client-side throttling when impersonating, making the kube-prom-stack upgrade take 25 minutes instead of one minute.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -5,7 +5,7 @@ go 1.17
 require (
 	github.com/fluxcd/pkg/apis/kustomize v0.3.1
 	github.com/fluxcd/pkg/apis/meta v0.10.2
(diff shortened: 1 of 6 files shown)


uderik:
After apply this patch, helmrelease reconcilation finished in 55.092283623s
with:
        - --kube-api-qps=1000
        - --kube-api-burst=1000

hiddeco (approved):
Made an attempt to move the growing set of constructor flags to options, but need to have a bit of a think about what they should be wired to (e.g. `rest.Config` or the `MemoryRESTClientGetter`). This looks good to me for now, thanks @stefanprodan and @uderik :bow:

PR #426, opened 2022-02-25
feat: add linux/s390x builds

skuethe:
Signed-off-by: skuethe <56306041+skuethe@users.noreply.github.com>

This PR adds the s390x platform to the build and release automation. As multi arch builds are setup perfectly already, the necessary changes are minimal.
A (stripped down) successful version of the github action can be seen here:
https://github.com/skuethe/helm-controller/actions/runs/1897491326

I have also just started a discussion about adding s390x to all flux components, just to not flood you with PRs without any feedback:
https://github.com/fluxcd/flux2/discussions/2475

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -31,6 +31,6 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./Dockerfile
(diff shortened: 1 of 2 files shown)


stefanprodan:
Closing this since we don't have a solution for libgit2.

PR #432, opened 2022-03-10
Support targeted Patches in the PostRenderer specification

trekawek:
This PR adds support for the `target` selector for strategic merge Kustomize patches, similarly as it's supported for the JSON6902 patches.

If the strategic merge patch have following form:
```yaml
- target:
    group: apps
    version: v1
    kind: Deployment
    name: nginx
  patch:
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: nginx
    spec:
      template:
        spec:
          containers:
            - name: nginx
              image: nignx:latest
```
it'll be translated into appropriate targeted [Patch](https://pkg.go.dev/github.com/fluxcd/pkg/apis/kustomize#Patch) in the effective Kustomize configuration.

Of course, the non-targeted form of the strategic merge patch is still supported:
```yaml
- apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: nginx
  spec:
    template:
      spec:
        containers:
          - name: nginx
            image: nignx:latest
```

The rationale behind it is a case in which certain Deployments or other K8s resources can be enabled or disabled with a Helm property. Now if we want to patch such a Deployment with the strategic merge patch (e.g. in order to add an env variable), the reconciliation will fail if the Deployment doesn't exist - this means we can't have a universal postRenderer configuration which will work regardless of the Helm values configuration.

The JSON6902 patches, with the explicit "target" section will work, as they'll simply ignore the missing targets. However, manipulating env variables with the JSON6902 patches is much harder than with the strategic merge patches and sometimes may not be possible. This is also true for other properties nested inside arrays, indexed by their "name" property (containers, envs, volumes, etc.)


hiddeco (commented):
Thanks for your contribution! :bow: :1st_place_medal: 

Given the kustomize-controller has a likewise feature, I think it would be better to align the APIs to work the same. This means that we would need the following:

1. The function from here: https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_generator.go#L220
2. A new `Patches` API field, as present in: https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta2/kustomization_types.go#L89. We import the same package (`github.com/fluxcd/pkg/apis/kustomize`), so this should be available.
3. Tiny bit of code to take the field into account:
   ```go
   for _, m := range kg.kustomization.Spec.Patches {
   	kus.Patches = append(kus.Patches, kustypes.Patch{
   		Patch:  m.Patch,
   		Target: adaptSelector(&m.Target),
   	})
   }
   ```

trekawek:
Thanks for the review @hiddeco! I applied your suggestions and adjusted the tests. Should I rebase/squash the commits to have a clean commit history for this feature?

hiddeco:
Yes, please rebase and squash :smile:

trekawek:
@hiddeco - done.

hiddeco (approved):
This looks good to me, thanks for following through on the suggested changes @trekawek :coffee:

trekawek:
That was quick. Thanks for the feedback and merging it so fast, much appreciated!

hiddeco:
The release will take a bit longer I am afraid, due to some extensive changes to the source-controller we are currently working on to finalize (which need to be taken into account on this side). If your patch is business critical, I can provide you with a RC build which needs to be combined with the updated CRD.

trekawek:
There's no need to create a RC build, but thanks for the offer @hiddeco!

PR #434, opened 2022-03-11
Update Helm to v3.8.1

hiddeco:
While moving away from the fork with a manual patch applied for memory
leak issues.


stefanprodan (approved):
LGTM

PS. We need this in SC also

hiddeco:
Had to make some changes to the fork for source-controller as it needs to be able to manage HTTP transport. PR is now pending in https://github.com/fluxcd/source-controller/pull/609.

PR #435, opened 2022-03-14
add fsgroup for securityContext

kingdonb:
Ref: https://github.com/fluxcd/flux2/issues/2537

In https://github.com/fluxcd/source-controller/issues/284#issuecomment-774703153 it seems that leaving fsGroup unset can result in this error.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -18,6 +18,10 @@ spec:
         prometheus.io/port: "8080"
     spec:
       terminationGracePeriodSeconds: 600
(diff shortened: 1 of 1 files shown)


hiddeco (commented):
Can you please add a comment as is done in the source-controller manifest? https://github.com/fluxcd/source-controller/blob/ce594e92ab60fb8c44cbe9c9155bd6e93f6faa44/config/manager/deployment.yaml#L23-L26

hiddeco (approved):

PR #443, opened 2022-03-21
Add Changelog note on new source-controller API v1beta2

pjbgf:


hiddeco (approved):

stefanprodan:
We need to same patch in KC changelog.

PR #452, opened 2022-03-28
Align version of dependencies when Fuzzing

ilanpillemer:
Ref: https://github.com/fluxcd/flux2/issues/2552

Signed-off-by: Ilan <ilan.pillemer@gmail.com>

@pjbgf as discussed in https://github.com/fluxcd/flux2/issues/2552


pjbgf (approved):
@ilanpillemer nice one! :bow:

LGTM

stefanprodan (approved):

PR #455, opened 2022-03-31
Add kubeconfig flags

pjbgf:
Two new flags were added to allow users to enable the use of user.Exec (`--insecure-kubeconfig-exec`) and InsecureTLS (`insecure-kubeconfig-tls`) in the kubeconfigs provided for remote apply.

Breaking change: both functionalities are no longer enabled by default.

Provides feature parity with `kustomize`: https://github.com/fluxcd/kustomize-controller/pull/593

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -24,6 +24,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/fluxcd/pkg/runtime/client"
 )
 
 func NewInClusterRESTClientGetter(cfg *rest.Config, namespace string) genericclioptions.RESTClientGetter {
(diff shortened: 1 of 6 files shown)


internal/kube/client.go
```diff
@@ -24,6 +24,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
+
+	fluxClient "github.com/fluxcd/pkg/runtime/client"
```
hiddeco, line 28: Any specific reason for this alias?

pjbgf: I was aligning it with the other reference in the project, but given that there is no other `/client` here, I removed that altogether.

stefanprodan (approved):
LGTM

Thanks @pjbgf 

PS. We need to document this in the Security section of our docs.

PR #461, opened 2022-04-11
Support defining a KubeConfig Secret data key

nstogner:
Fixes #458.

PR #472, opened 2022-04-29
kubeconfig: err on missing declared SecretRef.Key

hiddeco:
Follow up on #461 to align with changes suggested during review in https://github.com/fluxcd/kustomize-controller/pull/615.


stefanprodan (approved):
LGTM 

Thanks @hiddeco

PR #477, opened 2022-05-06
Prepare Helm release logic rewrite

hiddeco:
In short (see individual commits for further detail):

- Introduce a Helm storage observer to keep track of writes to the storage.
- Move HelmChart handling to a separate reconciler.
- Factor various bits out of the reconciler to slim it down to almost only Helm release related things.
- Move some elements to their own homes.

:warning: This is merged into `dev`, given the total size of the work. I am going to follow up with more PRs.


internal/kube/builder.go
```diff
@@ -0,0 +1,89 @@
...
+		opt(o)
+	}
+	if len(o.kubeConfig) > 0 {
+		return NewMemoryRESTClientGetter(o.kubeConfig, namespace, o.impersonateAccount, o.qps, o.burst, o.kubeConfigOptions)
+	}
+	if ia := o.impersonateAccount; ia != "" {
+		config := *config
+		SetImpersonationConfig(&config, namespace, ia)
+		return NewInClusterRESTClientGetter(&config, namespace)
+	}
```
pjbgf, line 87: I am not sure this accounts for the `DefaultServiceAccount` when `impersonateAccount` is empty.

internal/kube/builder_test.go
```diff
@@ -0,0 +1,94 @@
...
+		got := getter.(*MemoryRESTClientGetter)
+		g.Expect(got.namespace).To(Equal(namespace))
+		g.Expect(got.kubeConfig).To(Equal(cfg))
+		g.Expect(got.qps).To(Equal(qps))
+		g.Expect(got.burst).To(Equal(burst))
+		g.Expect(got.kubeConfigOpts).To(Equal(cfgOpts))
+		g.Expect(got.impersonateAccount).To(Equal(impersonate))
+	})
+
+	t.Run("with config and impersonate account", func(t *testing.T) {
```
pjbgf, line 80: We need an extra test case adding no impersonate account but it falls back to the `DefaultServiceAccount`.

pjbgf (approved):
LGTM

hiddeco:
Something is making the current end-to-end tests take precisely 5 minutes (and a little more) to be marked as Ready. I am unable to produce this locally however (using the same tests), so am inclined to say it has something to do with the bash magic in CI.

In any case, taking a note here, but not going to invest time in it now. As the tests are going to reworked anyway.

internal/storage/observer.go
```diff
@@ -0,0 +1,326 @@
...
+
+var (
+	// ErrReleaseNotObserved indicates the release has not been observed by
+	// the Observator.
+	ErrReleaseNotObserved = errors.New("release: not observed")
+)
+
+// Observator reports about the write actions to a driver.Driver, recorded as
+// ObservedRelease objects.
+type Observator interface {
```
darkowlzz, line 47: Maybe it'd be nice to add a line about the origin of the name.
I think it's based on the interfaces that the `driver.Driver` embeds Creator, Updator, Deletor, Queryor.

controllers/helmrelease_chart_controller.go
```diff
@@ -0,0 +1,273 @@
...
+				err = fmt.Errorf("failed to delete HelmChart '%s': %w", obj.Status.HelmChart, err)
+				return ctrl.Result{}, err
+			}
+			r.Eventf(obj, events.EventSeverityTrace, "HelmChartDeleted", "deleted HelmChart '%s'", obj.Status.HelmChart)
+		}
+		// Truncate the chart reference in the status object.
+		obj.Status.HelmChart = ""
+	}
+
+	if obj.DeletionTimestamp != nil {
```
darkowlzz, line 216: This seems like a new check in all the various `reconcileDelete()` implementations.
Is this necessary in certain situations or just added it here for safety?

hiddeco: It is new because `reconcileDelete` can happen both due to a change, and due to `Reconcile` calling it due to `DeletionTimestamp != nil`. Given this, it gate keeps a change from removing the finalizer.

internal/storage/observer.go
```diff
@@ -0,0 +1,326 @@
...
+	// GetObservedVersion returns the release with the given version if
+	// observed, or ErrReleaseNotObserved.
+	GetObservedVersion(name string, version int) (ObservedRelease, error)
+	// ObserveLastRelease observes the release in with the highest version in
+	// the embedded driver.Driver. It returns the driver.ErrReleaseNotFound is
+	// returned if a release with the provided name does not exist.
+	ObserveLastRelease(name string) (ObservedRelease, error)
+}
+
+// ObservedRelease is an immutable copy of a release.Release as observed to be
```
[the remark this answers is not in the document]
hiddeco: I have updated the doc blocs to clarify what I meant with "immutable", and also added a `DeepCopy` method.

darkowlzz (approved):
LGTM!

api/v2beta1/condition_types.go
```diff
@@ -16,6 +16,10 @@ limitations under the License.
 
 package v2beta1
 
+// ChartFinalizer is set on a HelmRelease when a HelmChart object is created
+// for it, and removed when this object has been deleted.
+const ChartFinalizer = "chart.finalizers.fluxcd.io"
```
stefanprodan, line 21: What's the role of this finalizer?

hiddeco: It is to prevent the HelmRelease from being removed while the HelmChart has not been cleaned up yet by the separate reocnciler.

stefanprodan: So if I do `flux delete hr` the HR will not be deleted not the Helm release uninstalled until SC GCs the HelmChart?

hiddeco: The Helm release will be uninstalled but the release will not be removed until the HelmChart has successfully been removed.

hiddeco: Note this is not different from the current logic, which does this as well https://github.com/fluxcd/helm-controller/blob/main/controllers/helmrelease_controller.go#L649-L652.

stefanprodan: > until the HelmChart has successfully been removed

Does it want for it to be removed from etcd or juat marked for deleting? Does this means HRs will hang indefinitely if SC fails to finalize the chart?

stefanprodan: Ok it just marks the chart for deletion instead of blocking. All good.

stefanprodan (approved):
LGTM

PR #480, opened 2022-05-10
kube: load KubeConfig (token) from FS on every reconcile

### LINKED ISSUE #479: Helm-controller pod is using stale tokens
Hi,

the "Bound Service Account Token Volume" is graduated to stable and enabled by default in Kubernetes version 1.22.
I am using "helm-controller:v0.21.0" in AWS EKS 1.22 and I have checked, if it is using stale tokens (regarding https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html and https://docs.aws.amazon.com/eks/latest/userguide/troubleshooting.html#troubleshooting-boundservicetoken).

So when the API server receives requests with tokens that are older than one hour, then it annotates the pod with "annotations.authentication.k8s.io/stale-token". In my case I can see the following annotation. E.g.:
`
"annotations":{"authentication.k8s.io/stale-token":"subject: system:serviceaccount:flux-system:helm-controller, seconds after warning threshold: 56187"`

Version:

helm-controller:v0.21.0

Cluster Details

AWS EKS 1.22

Steps to reproduce issue

- Enable EKS Audit Logs
- Query CW Insights (select cluster log group):

```
fields @timestamp
| filter @message like /seconds after warning threshold/
| parse @message "subject: *, seconds after warning threshold:*\"" as subject, elapsedtime
```

hiddeco:
4371610e4ba2b07e1e0c5641d1145ca3fb34ee07 is a partial cherry-pick of commit ae4f499, including
changes around `kube`. This to include some of the changes around the
construction of the ConfigFlags RESTClientGetter, as an attempt to
solve token refresh issues.

As icing on the cake, 93bd5ed64fd59a9f2078e32d5b29808ca3ba1a4a always retrieves the current token from the FS which fixes #479 and https://github.com/fluxcd/flux2/issues/2074.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -493,7 +493,6 @@ func (r *HelmReleaseReconciler) buildRESTClientGetter(ctx context.Context, hr v2
 		opts = append(opts, kube.WithKubeConfig(kubeConfig, r.Config.QPS, r.Config.Burst, r.KubeConfigOpts))
 	}
 	return kube.BuildClientGetter(r.Config, hr.GetReleaseNamespace(), opts...), nil
-
 }
 
 // composeValues attempts to resolve all v2beta1.ValuesReference resources
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -295,7 +295,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 	log := ctrl.LoggerFrom(ctx)
 
 	// Initialize Helm action runner
-	getter, err := r.getRESTClientGetter(ctx, hr)
+	getter, err := r.buildRESTClientGetter(ctx, hr)
 	if err != nil {
 		return v2.HelmReleaseNotReady(hr, v2.InitFailedReason, err.Error()), err
 	}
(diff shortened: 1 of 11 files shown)


pjbgf (approved):
LGTM

:top: :top:!

stefanprodan (approved):
I think we need to change the title and the description of this PR to be about token refresh. The code looks Ok to me.

hiddeco:
@stefanprodan I changed the PR title to make mention of it.

cgchinmay:
Hi, could you add some on how these changes fix the token refresh issue ?

hiddeco:
My understanding is that by initializing it from zero every time, it reads the latest runtime values and thereby is able to obtain a correct token.

Note that this part of the code is _very_ Helm specific, and the further controller (and controller-runtime manager) properly update their token automagically.

michael12312:
hi @hiddeco 

Per the doc from K8s mentioned:
The following Kubernetes client SDKs refresh tokens automatically within the required time frame:
Go v0.15.7 and later

I see helm-controller already use the version of k8s.io/client-go v0.23.6.

Per my understanding the expired token issue should not exists, can you explain why you put this fixing? Why load KubeConfig (token) from FS on every reconcile?

hiddeco:
As said before in https://github.com/fluxcd/helm-controller/pull/480#issuecomment-1126461691.

This patch is purely required because of how we integrate with Helm, and the controller-runtime client itself has been updating the token as expected. This process does however not update the token in the loaded REST config, from which we used to derive our client for Helm. It should not be taken as a further example on how to solve this for your project, as the further context around why and how we require this does likely not apply to you.

cgchinmay:
> As said before in [#480 (comment)](https://github.com/fluxcd/helm-controller/pull/480#issuecomment-1126461691).
> 
> This patch is purely required because of how we integrate with Helm, and the controller-runtime client itself has been updating the token as expected. This process does however not update the token in the loaded REST config, from which we used to derive our client for Helm. It should not be taken as a further example on how to solve this for your project, as the context around why and how we require this does likely not apply to you.

Thanks for the explanation. I had the same thought as @michael12312 to ask the question. Also, I noticed the same issue with [Multus](https://github.com/k8snetworkplumbingwg/multus-cni) even though it uses Go client > v0.15.7. So was trying to understand if there's a common pattern here.

hiddeco:
Think you need to look in the area of https://github.com/k8snetworkplumbingwg/multus-cni/blob/706de7c2c6a7d86401e1b0650bc050e4bdee984b/pkg/k8sclient/k8sclient.go#L384-L412 in that case. Although on a first 10 second glance, https://github.com/k8snetworkplumbingwg/multus-cni/blob/706de7c2c6a7d86401e1b0650bc050e4bdee984b/pkg/k8sclient/k8sclient.go#L405 should be doing the right thing and pretty much equals to what we are doing here.

hiddeco:
@michael12312 KubeCon and finalizing of source-controller changes got in the way for a bit. https://github.com/fluxcd/helm-controller/pull/491 is now pending.

PR #492, opened 2022-06-07
kube: configure proper account impersonation NS

hiddeco:
Fixing a regression introduced in #480 which would always pick the
namespace of the release. In addition, historically seen the
configuration of the impersonation username while making use of a
KubeConfig has never worked correctly, this has been adressed as well.


stefanprodan (approved):
LGTM

Thanks @hiddeco ☕

PR #503, opened 2022-07-01
Implement reconcile logic individual Helm actions

hiddeco:
This PR introduces architectural changes around the execution and observation model of Helm actions. In short:

- It introduces a new `v2beta2` API, with changes to the Status object. Deprecating `LastObserved*` and `LastApplied*` fields in favor of snapshot views of more detailed `HelmReleaseInfo` in the `.Current` and `.Previous` fields. The snapshot contains a digest of both the stored release object (which includes chart metadata), and values configuration.
- It introduces an `action` package which can run Helm actions based on the declared config in a `HelmRelease` object. Besides being of value to the controller, this could function as a way for the Flux CLI to provide more debug capabilities to a user.
- It introduces a `reconcile` package which contains `ActionReconciler` implementations, capable of running and observing individual Helm actions. Observation is done by injecting a `storage.ObserveFunc` into a Helm storage observer, which then takes note of writes to the storage during a Helm action in the Status of the HelmRelease object.
- It contains several other changes to earlier introduced changes in `dev`, to allow the consumer of the storage observer to inject their own API instead of keeping track of state itself. The controller does this using the newly introduced `release` package, which contains the earlier in storage embedded observation bits and a few utilities to translate this to a `HelmReleaseInfo` object, etc.
- It introduces an outline for the decision making model, which is by no means perfect at present. But sufficient to use as a foundation for the detailed implementation.
- It introduces a `digest` package which allows for configuring a canonical digest algorithm at runtime. This will allow users to e.g. configure BLAKE3 which is much faster on most modern hardware, but can't be set as a default due to it not being an industry standard (yet).
- It has (especially compared to current `main`) tests, loads of tests.

The API is still subject to changes, and the backwards compatibility model hasn't been decided on yet. In addition, it does not contain glue to step through sub-reconcile actions, this is the next follow-up.


api/v2beta2/groupversion_info.go
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright 2020 The Flux authors
```
pjbgf, line 2: Across the new files:
```suggestion
Copyright 2022 The Flux authors
```

internal/release/digest.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"encoding/json"
+
+	godigest "github.com/opencontainers/go-digest"
+)
+
+// Digest calculates the digest of the given ObservedRelease by JSON encoding
+// it into a hash.Hash of the given digest.Algorithm. The algorithm is expected
+// to have been confirmed to be available by the caller, not doing this may
+// result in panics.
```
pjbgf, line 28: Would it be worth abstracting this away from our callers, by recovering potential panics and returning an error instead?

hiddeco: No, as this is all within our own control.

api/v2beta2/helmrelease_types.go
```diff
@@ -0,0 +1,1036 @@
...
+
+	// Previous holds the latest observed HelmReleaseInfo for the previous
+	// release.
+	// +optional
+	Previous *HelmReleaseInfo `json:"previous,omitempty"`
+
+	// Failures is the reconciliation failure count against the latest desired
+	// state. It is reset after a successful reconciliation.
+	// +optional
+	Failures int64 `json:"failures,omitempty"`
```
pjbgf, line 876: Can we change this to `int` so it is less platform-specific? Otherwise we will have to remove support to `linux/armv7`.

Running this version on that platform results in:
`standard_init_linux.go:228: exec user process caused: exec format error`

internal/reconcile/unlock.go
```diff
@@ -0,0 +1,100 @@
...
+	// Ensure latest is still same as current.
+	obs := release.ObserveRelease(rls)
+	if obs.Targets(cur.Name, cur.Namespace, cur.Version) {
+		if status := rls.Info.Status; status.IsPending() {
+			// Update pending status to failed and persist.
+			rls.SetStatus(helmrelease.StatusFailed, fmt.Sprintf("Release unlocked from stale '%s' state",
+				status.String()))
+			if err = cfg.Releases.Update(rls); err != nil {
+				req.Object.Status.Failures++
+				conditions.MarkFalse(req.Object, helmv2.ReleasedCondition, "StalePending",
```
pjbgf, line 69: `"StalePending"` could be converted into a new const `helmv2.StalePendingCondition`.

api/v2beta2/helmrelease_types.go
hiddeco: This is copied over from the v2beta1 API...

internal/reconcile/unlock.go
hiddeco: It is part of the release cycle state and not a separate condition.

internal/action/crds.go
```diff
@@ -0,0 +1,190 @@
...
+				})
+			} else if !apierrors.IsNotFound(err) {
+				cfg.Log("failed to get CRD %s: %s", r.Name, err)
+				return err
+			}
+		}
+		// Send them to Kube
+		if rr, err := cfg.KubeClient.Update(original, allCRDs, true); err != nil {
+			cfg.Log("failed to apply CRD %s", err)
+			return errors.New(fmt.Sprintf("failed to apply CRD %s", err))
```
souleb, line 155:
```suggestion
			return fmt.Errorf("failed to apply CRD %s", err)
```

internal/action/crds.go
```diff
@@ -0,0 +1,190 @@
...
+				}
+				if rr.Updated != nil {
+					totalItems = append(totalItems, rr.Updated...)
+				}
+				if rr.Deleted != nil {
+					totalItems = append(totalItems, rr.Deleted...)
+				}
+			}
+		}
+		break
```
souleb, line 169:
```suggestion
```

api/v2beta2/helmrelease_types.go
```diff
@@ -0,0 +1,1036 @@
...
+
+const (
+	// Skip CRDs do neither install nor replace (update) any CRDs.
+	Skip CRDsPolicy = "Skip"
+	// Create CRDs which do not already exist, do not replace (update) already existing
+	// CRDs and keep (do not delete) CRDs which no longer exist in the current release.
+	Create CRDsPolicy = "Create"
+	// Create CRDs which do not already exist, Replace (update) already existing CRDs
+	// and keep (do not delete) CRDs which no longer exist in the current release.
+	CreateReplace CRDsPolicy = "CreateReplace"
```
aryan9600, line 500: maybe?
```suggestion
	CreateAndReplace CRDsPolicy = "CreateAndReplace"
```

api/v2beta2/helmrelease_types.go
```diff
@@ -0,0 +1,1036 @@
...
+func (in Uninstall) GetTimeout(defaultTimeout metav1.Duration) metav1.Duration {
+	if in.Timeout == nil {
+		return defaultTimeout
+	}
+	return *in.Timeout
+}
+
+// HelmReleaseInfo holds the status information for a Helm release as performed
+// by the controller.
+type HelmReleaseInfo struct {
```
aryan9600, line 790: Maybe it'd be helpful for this to have a `StorageNamespace` field as well.

internal/action/install.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return install, nil
+}
+
+func installCRDs(config *helmaction.Configuration, obj *helmv2.HelmRelease, chrt *helmchart.Chart, install *helmaction.Install) error {
+	policy, err := crdPolicyOrDefault(obj.Spec.GetInstall().CRDs)
+	if err != nil {
+		return err
+	}
+	if policy == helmv2.Skip || policy == helmv2.CreateReplace {
```
souleb, line 88: is that correct to skip crds if policy is `helmv2.CreateReplace`?

internal/action/log_test.go
```diff
@@ -0,0 +1,103 @@
...
+		want      string
+	}{
+		{name: "log", size: 2, fill: []string{"a", "b", "c"}, wantCount: 3, want: "b\nc"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var count int
+			l := NewLogBuffer(func(format string, v ...interface{}) {
+				count++
+				return
```
souleb, line 40:
```suggestion
```

internal/action/upgrade.go
```diff
@@ -0,0 +1,94 @@
...
+	renderer, err := postrender.BuildPostRenderers(rel)
+	if err != nil {
+		return nil, err
+	}
+	upgrade.PostRenderer = renderer
+
+	return upgrade, err
+}
+
+func upgradeCRDs(config *helmaction.Configuration, obj *helmv2.HelmRelease, chrt *helmchart.Chart) error {
```
souleb, line 80: should we check 'helmv2.Install` and set `upgrade.SkipCRDs = true` accordingly?

api/v2beta2/helmrelease_types.go
```diff
@@ -0,0 +1,1036 @@
...
+	ChartName string `json:"chartName"`
+	// ChartVersion is the chart version of the release object in
+	// storage.
+	// +required
+	ChartVersion string `json:"chartVersion"`
+	// ConfigDigest is the checksum of the config (better known as
+	// "values") of the release object in storage.
+	// It has the format of `<algo>:<checksum>`.
+	// +required
+	ConfigDigest string `json:"configDigest"`
```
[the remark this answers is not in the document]
hiddeco: This matches the name of the field it is matched against from the Helm API.

api/v2beta2/helmrelease_types.go
hiddeco: This will be added when the logic for changes to the `targetNamespace` and/or `storageNamespace` is added to the release, which includes e.g. garbage collection.

api/v2beta2/helmrelease_types.go
hiddeco: This is copied over from the previous API version, and not worth the change.

internal/action/install.go
hiddeco: I think this is inherited from the current code in `runner`, but will double check when I address some of the other suggestions.

internal/action/upgrade.go
hiddeco: Skip CRDs is only taken into account for install, see current logic in `runner`. The `SkipCRDs` flag for `Upgrade` from Helm does only apply when it is combined with `Upgrade.Install`, which is not a viable configuration option within our context.

api/v2beta2/helmrelease_types.go
pjbgf: That's right, the issue with `exec format error` was caused when building it locally on a broken buildx setup.

internal/release/observed.go
```diff
@@ -0,0 +1,200 @@
...
+// Targets returns if the release matches the given name, namespace and
+// version. If the version is 0, it matches any version.
+func (o ObservedRelease) Targets(name, namespace string, version int) bool {
+	return o.Name == name && o.Namespace == namespace && (version == 0 || o.Version == version)
+}
+
+// Encode JSON encodes the ObservedRelease and writes it into the given writer.
+func (o ObservedRelease) Encode(w io.Writer) error {
+	enc := json.NewEncoder(w)
+	if err := enc.Encode(o); err != nil {
```
souleb, line 97: It is stated above that the `Manifest` is omitted in JSON representation, but this encode the whole object.

internal/release/observed.go
```diff
@@ -0,0 +1,200 @@
...
+		Version:   rel.Version,
+		Config:    nil,
+		Manifest:  rel.Manifest,
+		Hooks:     nil,
+		Namespace: rel.Namespace,
+		Labels:    nil,
+	}
+
+	if rel.Info != nil {
+		obsRel.Info = *rel.Info
```
souleb, line 127: is `rel.Info` immutable? Why are we not copying it like the other fields?

internal/action/config.go
```diff
@@ -0,0 +1,162 @@
...
+	return func(f *ConfigFactory) error {
+		if f.KubeClient == nil {
+			return fmt.Errorf("failed to set debug log: no Kubernetes client configured")
+		}
+		f.KubeClient.Log = LogV(log, logger.DebugLevel)
+		return nil
+	}
+}
+
+// Build returns a new Helm action.ConfigFactory configured with the
```
souleb, line 133:
```suggestion
// Build returns a new Helm action.Configuration configured with the
```

internal/release/observed.go
hiddeco: Because unlike all other structures, this structure is simple and can simply copy the values: https://github.com/helm/helm/blob/main/pkg/release/info.go#L23

relu (commented):
Added some minor comments, otherwise this looks really nice.

internal/reconcile/action.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+		}
+	}
+	return nil, ErrReconcileEnd
+}
+
+func rollbackOrUninstall(factory *action.ConfigFactory, req *Request) (ActionReconciler, error) {
+	remediation := req.Object.Spec.GetInstall().GetRemediation()
+	if req.Object.Status.Previous != nil {
+		// TODO: determine if previous still in storage and unmodified
```
pjbgf, line 100:
```suggestion
		// TODO: determine if previous is still in storage and unmodified
```

internal/release/observed.go
```diff
@@ -0,0 +1,200 @@
...
+	// ChartMetadata contains the current Chartfile data of the release.
+	ChartMetadata chart.Metadata `json:"chartMetadata"`
+	// Config is the set of extra Values added to the chart.
+	// These values override the default values inside the chart.
+	Config map[string]interface{} `json:"config"`
+	// Manifest is the string representation of the rendered template.
+	// It is omitted from the JSON representation of the object for performance
+	// reasons. ManifestChecksum can be used to notice changes, after which
+	// Manifest can be inspected while loaded.
+	Manifest string `json:"manifest"`
```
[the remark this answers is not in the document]
hiddeco: The comment is actually incorrect, performance is no longer an issue with the changes to the observer and the manifest is included.

internal/release/observed.go
hiddeco: See https://github.com/fluxcd/helm-controller/pull/503#discussion_r914676584

internal/reconcile/install.go
```diff
@@ -0,0 +1,74 @@
...
+
+		// Return error if we did not store a release, as this does not
+		// require remediation and the caller should e.g. retry.
+		if newCur := req.Object.Status.Current; newCur == nil || cur == newCur {
+			return err
+		}
+
+		// Count install failure on object, this is used to determine if
+		// we should retry the install and/or remediation.
+		req.Object.Status.InstallFailures++
```
pjbgf, line 57: Do you mind clarifying why an error on `action.Install` would lead to incrementing `req.Object.Status.Failures` but not `req.Object.Status.InstallFailures` - when the release is not stored.

internal/action/install.go
hiddeco: Matches current: https://github.com/fluxcd/helm-controller/blob/main/internal/runner/runner.go#L121-L123

internal/reconcile/action_test.go
```diff
@@ -0,0 +1,496 @@
...
+					Current: release.ObservedToInfo(release.ObserveRelease(releases[0])),
+				}
+			},
+			chart:  testutil.BuildChart(),
+			values: map[string]interface{}{"bar": "foo"},
+			want:   &Upgrade{},
+		},
+		// {
+		//	name: "manifestTmpl diff requires upgrade (or apply?) when enabled",
+		// },
```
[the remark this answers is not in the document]
pjbgf: Maybe add a `//TODO:` to make that clearer?

internal/reconcile/install.go
hiddeco: The `<Install|Upgrade>Failures` are used to determine if we should stop after an `x` number of retries, and is also related to remediation. This stop is mostly present to not cause immense drift to the Helm storage (in terms of `version++`).

Given this, and the fact that  remediation is not required if there is no new object in the storage (because Helm did not apply anything, as this is a prerequisite of it applying things). Anything that does not touch the storage is better retried (indefinitely), than it to stop and wait for a human, and thus not counted as a "full attempt".

internal/reconcile/action_test.go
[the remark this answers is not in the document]
hiddeco: I feel this is already captured in the various `// TODOs` in the actual `action.go` code.

internal/reconcile/install.go
pjbgf: That's great context, thank you. I think the part around the motivation for this would be good to also be captured in that comment - specially the part around retrying indefinitely.

internal/reconcile/release.go
```diff
@@ -0,0 +1,65 @@
...
+	// This can happen for actions where targeting a release by version is not
+	// possible, for example while running tests.
+	ErrReleaseMismatch = errors.New("release mismatch")
+)
+
+// observeRelease returns a storage.ObserveFunc which updates the Status.Current
+// and Status.Previous fields of the HelmRelease object. It can be used to
+// record Helm install and upgrade actions as - and while - they are written to
+// the Helm storage.
+func observeRelease(obj *helmv2.HelmRelease) storage.ObserveFunc {
```
souleb, line 47: the naming is similar to ObserveRelease. Maybe a better naming for this function that returns an `ObserveFunc` would be `observeReleaseFunc`?

internal/reconcile/rollback.go
```diff
@@ -0,0 +1,94 @@
...
+		// Mark failure on object.
+		req.Object.Status.Failures++
+		conditions.MarkFalse(req.Object, helmv2.RemediatedCondition, helmv2.RollbackFailedReason, err.Error())
+
+		// Return error if we did not store a release, as this does not
+		// affect state and the caller should e.g. retry.
+		if newCur := req.Object.Status.Current; newCur == nil || newCur == cur {
+			return err
+		}
+		return nil
```
souleb, line 69: could you explain why do we silently return here?

internal/reconcile/rollback.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (r *Rollback) Name() string {
+	return "rollback"
+}
+
+func (r *Rollback) Type() ReconcilerType {
+	return ReconcilerTypeRemediate
+}
+
+func (r *Rollback) Reconcile(ctx context.Context, req *Request) error {
```
souleb, line 47: Missing comment.

internal/release/observed.go
souleb: the filter option seems to be used only in the tests. Is there another use case? Otherwise maybe we can have a specific unexported func for filtering.

internal/release/observed.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		rel.Hooks = hooks
+	}
+}
+
+// ObservedRelease is a copy of a Helm release object, as observed to be written
+// to the storage by a storage.Observer. The object is detached from the Helm
+// storage object, and mutations to it do not change the underlying release
+// object.
+type ObservedRelease struct {
```
souleb, line 62: After going through the code, I haven't seen this object mutated. Will this happen in the reconciler?

internal/release/observed.go
hiddeco: It always uses the default filters, this might change depending on what actually looping through the results produces with current code. Which is why I made it part of the official API.

internal/release/observed.go
hiddeco: We do not, it's noted for documentation purposes.

internal/reconcile/release.go
hiddeco: Why? `ObserveRelease` comes from another package so is always prefixed. None of the other functions are suffixed with `func`.

internal/reconcile/rollback.go
hiddeco: Yes, not planning to add them at this stage.

internal/action/verify.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		if errors.Is(err, helmdriver.ErrReleaseNotFound) {
+			if curRel != nil && curRel.Name == obj.GetReleaseName() && curRel.Namespace == obj.GetReleaseNamespace() {
+				return nil, ErrReleaseDisappeared
+			}
+			return nil, ErrReleaseNotFound
+		}
+		return nil, err
+	}
+	if curRel == nil {
```
souleb, line 60: shouldn't we check this before checking if a release has disappeared?

hiddeco: No, because for a release not to be observed it has to exist. So we do everything that depends on the error (which also confirms existence) first.

darkowlzz (approved):
LGTM!

internal/action/install.go
```diff
@@ -54,7 +55,7 @@ func Install(ctx context.Context, config *helmaction.Configuration, obj *helmv2.
 	return install.RunWithContext(ctx, chrt, vals.AsMap())
 }
 
-func newInstall(config *helmaction.Configuration, obj *helmv2.HelmRelease) (*helmaction.Install, error) {
+func newInstall(config *helmaction.Configuration, obj *helmv2.HelmRelease, opts []InstallOption) *helmaction.Install {
```
souleb, line 58: we could have the function like:

```go
func newInstall(config *helmaction.Configuration, obj *helmv2.HelmRelease, opts ...InstallOption) *helmaction.Install {...}
```

Like this, you don't have to pass a `nil` parameter when you don't have any option.

pjbgf (approved):
LGTM!

great job @hiddeco! 🥷

souleb (approved):
LGTM! Thanks for answering all the questions 🙇

PR #511, opened 2022-07-26
Adding test filters

jtyr:
This PR is adding support for test filters.


hiddeco:
Thanks for the contribution @jtyr! I recently noticed as well that this seems to be a missing feature.

However, while I am willing to accept this feature request, it is strategically better for me if this would be added to the `dev` branch which contains the outline for the new `v2beta2` API (and _a lot_ of new code around running Helm actions).

Given this, is this feature pressing for you, or could it wait for a little longer?

jtyr:
This feature is not immediately required from our team. I can create the PR for the `dev` branch if you would prefer that. How long do you think it will take to release changes from the `dev` branch?

internal/runner/runner.go
```diff
@@ -349,6 +349,20 @@ func (r *Runner) Test(hr v2.HelmRelease) (*release.Release, error) {
 	test.Namespace = hr.GetReleaseNamespace()
 	test.Timeout = hr.Spec.GetTest().GetTimeout(hr.GetTimeout()).Duration
 
+	filters := make(map[string][]string)
```
hiddeco, line 352: In the `dev` branch the `runner` will be defunct once I finalize the wiring of the `internal/reconcile` package with the bits in `controllers` (which is the chunk of work I am chewing on at present). Given this, it would be better to move this to `internal/reconcile/test.go` which should be straight forward (and also allows for easier adding of tests :-)).

api/v2beta1/helmrelease_types.go
```diff
@@ -691,6 +691,9 @@ type Test struct {
 	// actions in 'Install.IgnoreTestFailures' and 'Upgrade.IgnoreTestFailures'.
```
hiddeco, line 691: This API will be deprecated and changes to `v2beta1` can be omitted.

hiddeco:
@jtyr our focus for next minor Flux release is landing support for `OCIRepository` objects. After this the helm-controller changes are up next, which means the timeline you are looking at is realistically end of August / beginning of September.

api/v2beta1/helmrelease_types.go
jtyr: Done

internal/runner/runner.go
jtyr: You probably meant to move it to `internal/action/test.go`. That's done now.

hiddeco: Oops, that's indeed what I meant.

api/v2beta2/helmrelease_types.go
```diff
@@ -706,6 +709,24 @@ func (in Test) GetTimeout(defaultTimeout metav1.Duration) metav1.Duration {
 	return *in.Timeout
 }
 
+// Filters holds the configuration for individual Helm test filters.
+type Filter struct {
+	// Name is the name of the test.
+	Name string `json:"name"`
+	// Exclude is specifies wheter the named test should be excluded.
```
hiddeco, line 716:
```suggestion
	// Exclude is specifies whether the named test should be excluded.
```

jtyr:
@hiddeco Is there anything else needed to get this merged into the `dev` branch?

hiddeco (approved):
It was just waiting for me to be back from time off. This looks good to me, thanks @jtyr :bow:

PR #513, opened 2022-08-08
Disable caching of secrets and configmaps

### LINKED ISSUE #512: Flux unnecessarily fetching all secrets in the cluster
I've noticed that when you specify `valuesFrom` in a HelmRelease, the helm-controller won't just fetch that one secret; it will `list` all secrets in the whole cluster instead:

```
W0802 16:18:16.709401       7 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.1/tools/cache/reflector.go:167: 
  failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:flux-system:helm-controller" 
  cannot list resource "secrets" in API group "" at the cluster scope
```

I understand Flux is meant to be installed cluster-wide on the assumption that you can use Flux to install stuff in any namespace, but I don't have any intention of doing that in e.g. "kube-system", so I've deployed flux with custom permissions that only allow it to read secrets in specific namespaces. So that `list` operation fails for me.

I believe the `list` operation is triggered because of the client-go caching mechanism. The helm-controller's code doesn't perform a list: 
https://github.com/fluxcd/helm-controller/blob/5a7d0f8535406d155ed7d0634b27574538febed0/controllers/helmrelease_controller.go#L556
But maybe the cache config of the controller-runtime client may have the answer? https://github.com/kubernetes-sigs/controller-runtime/blob/f0351217e9e026533aece74d054b23cae5441659/pkg/client/client.go#L79

mac-chaffee:
Fixes #512 .

By default, `controller-runtime` gives you a cache-backed client, which will perform a LIST operation on any API type that you fetch, even if you use a GET request like we do: https://github.com/fluxcd/helm-controller/blob/5a7d0f8535406d155ed7d0634b27574538febed0/controllers/helmrelease_controller.go#L556

So unless you restrict helm-controller to a specific namespace with `--watch-all-namespaces=false`, then it will attempt to populate its cache by listing all secrets in the cluster, even if it only needs to access a few in specific namespaces. That requires giving helm-controller the ability to read every secret in the cluster.

This PR allows users to avoid that restriction by disabling caching of secrets. Now users can deploy HelmReleases that use `valuesFrom` without having to grant Flux access to more secrets than necessary.

### DIFF
--- a/internal/features/features.go
+++ b/internal/features/features.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package features sets the feature gates that
+// helm-controller supports, and their default states.
+package features
+
+import feathelper "github.com/fluxcd/pkg/runtime/features"
+
+const (
+	// CacheSecretsAndConfigMaps configures the caching of Secrets and ConfigMaps
+	// by the controller-runtime client.
+	//
+	// When enabled, it will cache both object types, resulting in increased memory usage
+	// and cluster-wide RBAC permissions (list and watch).
+	CacheSecretsAndConfigMaps = "CacheSecretsAndConfigMaps"
+)
+
+var features = map[string]bool{
+	// CacheSecretsAndConfigMaps
+	// opt-in from v0.28
+	CacheSecretsAndConfigMaps: false,
+}
+
+// FeatureGates contains a list of all supported feature gates and
+// their default values.
+func FeatureGates() map[string]bool {
+	return features
+}
+
+// Enabled verifies whether the feature is enabled or not.
+//
+// This is only a wrapper around the Enabled func in
+// pkg/runtime/features, so callers won't need to import
+// both packages for checking whether a feature is enabled.
+func Enabled(feature string) (bool, error) {
+	return feathelper.Enabled(feature)
+}
+
+// Disable disables the specified feature. If the feature is not
+// present, it's a no-op.
+func Disable(feature string) {
+	if _, ok := features[feature]; ok {
+		features[feature] = false
+	}
+}
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	crtlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	"github.com/fluxcd/pkg/runtime/acl"
@@ -40,10 +41,13 @@ import (
 	"github.com/fluxcd/pkg/runtime/pprof"
 	"github.com/fluxcd/pkg/runtime/probes"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	corev1 "k8s.io/api/core/v1"
 
 	v2 "github.com/fluxcd/helm-controller/api/v2beta1"
 	"github.com/fluxcd/helm-controller/controllers"
+	"github.com/fluxcd/helm-controller/internal/features"
 	intkube "github.com/fluxcd/helm-controller/internal/kube"
+	feathelper "github.com/fluxcd/pkg/runtime/features"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -74,6 +78,7 @@ func main() {
 		httpRetry               int
 		clientOptions           client.Options
 		kubeConfigOpts          client.KubeConfigOptions
+		featureGates            feathelper.FeatureGates
 		logOptions              logger.Options
 		aclOptions              acl.Options
 		leaderElectionOptions   leaderelection.Options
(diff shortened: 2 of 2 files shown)


pjbgf:
Overall, I do see the benefits of offering as an option the capability of decreasing the RBAC requirements in which the controller runs under. As discussed [here](https://github.com/fluxcd/helm-controller/issues/512#issuecomment-1207074434) even when impersonation is in place, the controller service account still requires `list` access at cluster-level. 

There are some considerations worth taking around the impact of this. But giving Flux users the option to decide based on their priorities and security vs performance trade-offs, does not seem like a terrible idea.

I am just not sure about the API. Potentially, users may want to disable cache for other objects in the future, so `--disable-cache-for` may more flexible on that vein.  But, I would be keen on hearing other maintainers take on this.

mac-chaffee:
@pjbgf I was thinking the same thing originally about `--disable-cache-for`, but then we'd have to parse strings like "core/v1/Secrets" into Kubernetes objects, which didn't seem straightforward to me (although I am a weak golang coder 😅 )

If you know of a helper function that does the job, I'd be happy to include it!

mac-chaffee:
Hello again! Just following up on this. Given the complexity of parsing API types from strings and the lack of a defined use-case for disabling caches for things other than secrets, I think we ought to stick with just `--no-cache-secrets`

pjbgf:
@mac-chaffee after some thought on this and discussions with @darkowlzz, we think that `ConfigMaps` and `Secrets` could be tied in together, as this can lower the running memory requirements, on top of the RBAC requirements benefits you are looking for.

This has been done in [CAPI](https://github.com/kubernetes-sigs/cluster-api/commit/1ac789696f38282d236e0318221fb95753cbddbb) a while ago and should work well for us too.

Therefore, can you please:

1) Bundle both `ConfigMaps` and `Secrets` as part of this PR.
2) Add an **opt-out** Feature Gate instead of a flag: `CacheSecretsAndConfigMaps`. Source controller is a good example for feature gates in Flux: https://github.com/fluxcd/source-controller/blob/885c9f2cba44bb8a9130ef839a1dfcf0bead2c94/internal/features/features.go#L35-L37.
3) Rebase the PR.

mac-chaffee:
Sounds good. To clarify, should CacheSecretsAndConfigMaps be set to true by default? or false?

internal/features/features.go
```diff
@@ -0,0 +1,56 @@
...
+// helm-controller supports, and their default states.
+package features
+
+import feathelper "github.com/fluxcd/pkg/runtime/features"
+
+const (
+	// When enabled, CacheSecretsAndConfigMaps will allow the controller-runtime
+	// client to exhibit its default behavior of caching all resources that helm-controller
+	// requests. Since this may cause high memory usage and require cluster-wide list
+	// and watch permissions, it is disabled by default.
```
pjbgf, line 27:
```suggestion
	// CacheSecretsAndConfigMaps configures the caching of Secrets and ConfigMaps
	// by the controller-runtime client.
	// 
	// When enabled, it will cache both object types, resulting in increased memory usage
	// and cluster-wide RBAC permissions (list and watch).
```

internal/features/features.go
```diff
@@ -0,0 +1,56 @@
...
+	CacheSecretsAndConfigMaps = "CacheSecretsAndConfigMaps"
+)
+
+var features = map[string]bool{
+	CacheSecretsAndConfigMaps: false,
```
pjbgf, line 32:
```suggestion
	// CacheSecretsAndConfigMaps
	// opt-in from v0.28
	CacheSecretsAndConfigMaps: false,
```

pjbgf (changes requested):
@mac-chaffee please align the documentation to the other controllers (as per comments below).

pjbgf:
> To clarify, should CacheSecretsAndConfigMaps be set to true by default? or false?

`false` by default looks good, this will help us understand the impact more quickly whilst also providing a way out for users.
I am running a few tests with this version. Can you confirm the RBAC needs aligns with the expectations (as per linked issue)?

mac-chaffee:
> Can you confirm the RBAC needs aligns with the expectations (as per linked issue)?

Yes I ran it locally and was unpleasantly surprised to see it connect to my production cluster, but at least I was able to confirm the RBAC was working with the flag enabled/disabled!

pjbgf:
Tested this against my setup, all works as expected. In terms of memory footprint, I don't have many secrets and configmaps in the cluster (~105 objects), but it already makes a dent. At a larger cluster this should be even more substantial:

|        | Cache Enabled | Cache Disabled |
|--------|---------------|----------------|
| Peak   |     112MiB    |      66MiB     |
| Valley |     43MiB     |      17MiB     |

pjbgf (approved):
LGTM

But would be good to get other maintainer's view too.

souleb:
I really like this addition thanks @mac-chaffee. I agree that this should be added to the other controllers behind feature gate.

hiddeco:
To sort the issue in the e2e tests, please run `make fmt vet`. Issue with the fuzz tests is sorted by rebasing.

mac-chaffee:
Oh wait I forgot the docs. Seems we don't have docs for the existing CLI flags, so I can add that too. Thinking the most logical place to put that would be https://github.com/fluxcd/helm-controller/blob/main/docs/spec/README.md instead of the page on helmreleases since spec/README is linked from the top-level README as "controller".

hiddeco:
@mac-chaffee we do have flag documentation, but they're maintained in the `options.md` files in the directories found here: https://github.com/fluxcd/website/tree/main/content/en/flux/components

souleb:
Is it really fine to have it `false` by default? This changes the controller behaviour so I would and I wonder what is the impact on a cluster where everything is deployed with Flux.

mac-chaffee:
> @mac-chaffee we do have flag documentation, but they're maintained in the `options.md` files in the directories found here: https://github.com/fluxcd/website/tree/main/content/en/flux/components

Ah thanks. I removed the docs from this PR (we don't want two sources of truth for the cli flags) and will update fluxcd/website later

main.go
```diff
@@ -96,10 +101,19 @@ func main() {
...
 	rateLimiterOptions.BindFlags(flag.CommandLine)
 	kubeConfigOpts.BindFlags(flag.CommandLine)
+	featureGates.BindFlags(flag.CommandLine)
 	flag.Parse()
 
 	ctrl.SetLogger(logger.NewLogger(logOptions))
 
+	err := featureGates.WithLogger(setupLog).
+		SupportedFeatures(features.FeatureGates())
+
```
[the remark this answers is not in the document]
mac-chaffee: Turns out we do need that line to load the features: https://github.com/fluxcd/helm-controller/actions/runs/3720709054/jobs/6326220127#step:44:79

hiddeco:
> Is it really fine to have it `false` by default? This changes the controller behaviour so I would and I wonder what is the impact on a cluster where everything is deployed with Flux.

Think based on the (realistic) assumption that we only need a tiny subset of Secrets in any cluster makes this a fair choice. Helm itself maintains its own (extremely expensive) cache anyway, so in terms of API request overhead this should be negligible.

hiddeco (approved):
We discussed this among maintainers and agreed this would be beneficial to all controllers. We would like the feature flag (`CacheSecretsAndConfigMaps`) to be the same across controllers (even if they do not necessarily make use of a `ConfigMap` at present). For which I will create an umbrella issue.

Thanks a lot @mac-chaffee :1st_place_medal:

Issue #412, opened 2022-02-10
Customizable events


lictw:
Hi! It will be great if i can flexibly customize alerts that i receiving in Slack.

What missing now:
1. Ability to disable event about start of upgrade.
2. Option to enable printing diff of prev./current values (it's about what changed, what reconciled).
3. Ability to send (or at least collect somewhere) helm logs to see depreciation warnings.
4. Final message as customized Golang template.

As i understand your architecture - HC (current project) generates *Event for NC (notification-controller), NC sends events, so:
1. Option in HelmRelease CRD - don't generate some event.
2. Option in HelmRelease CRD - apply diff as [metadata of event](https://github.com/fluxcd/pkg/blob/main/runtime/events/event.go#L69).
3. Same as above.
4. It's for notification-controller.

ArtunSubasi:
I would also love to have this feature. Currently, slack notifications look like this:

**helmrelease/my-service**
Helm upgrade has started
**revision**
3.1.0
**summary**
Some static text

It would be great if we can insert some placeholders (such as certain helm parameter values) either to the summary or to the event message.

Issue #256, opened 2021-04-22
Enrich helm events with user specified attributes to make notifications more informational


willejs:
Currently, helm release events just contain the helm chart version, which is not particularly useful as in most cases the underlying image tag is changing, but not the chart version. In your deployments slack channel, you will probably see something like this:
![Screenshot 2021-04-22 at 10 48 56](https://user-images.githubusercontent.com/1313453/115694047-690f6c00-a358-11eb-83c2-70d99a53cc43.png)
What I really want to know is probably which image tag is getting deployed, perhaps I could specify a go template string and use either git commit info from the original control repo, or values in the supplied values.yaml to identify the image tag?

PR #514, opened 2022-08-10
Include chart annotations as event metadata

eljulians:
This PR extends the Helm release controller to include Chart `annotations` in the registered event as metadata, along with the Chart revision already included.

This is my first Go code, so feel free to give opinionated feedback :)

# Rationale

## Use case
With the current notifications, is hard to be aware of what exactly was deployed, as just the Helm chart revision is included in the payload. If I wanted to know what specific change (or changeset) has been rolled out, it wouldn't be possible with the current setup. A workaround could be to abuse the chart `version` semver, but of course with several drawbacks, like needing to keep a 1-1 relationship between the char and app versions, having to come up with some specific encoding, having it to decode on the other end if a generic webhook receiver has been configured, and just probably being a bad practice.

It's probably reasonable to be able to plug some arbitrary data into the event delivered by Flux, specially considering that the Helm charts already provide annotations for this. I had a look into the open issues and it's something that was mentioned several times (https://github.com/fluxcd/helm-controller/issues/412 and https://github.com/fluxcd/helm-controller/issues/256 just to mention a couple of examples).

By including the chart annotations as part of the metadata, users can enrich their notifications as they wish by including the data they consider necessary for their own use cases.

Doing it with the chart annotations, the user experience doesn't change, as the chart needs to be updated for making a release anyways, and the data can be set at that point; or just left it empty otherwise if it's not needed, as they're optional.

## Changes

The calls to the `event` function from `reconcile` are done passing `nil` as the chart is not loaded yet at that point.

The annotations are merged in the root of the `event` map instead of having a nested structure, for simplicity, e.g. the payload looks like:

```json
{
    "revision": "1.0.2",
    "annotation_key1": "annotation_key2"
}
```

Instead of

```json
{
    "revision": "1.0.2",
    "annotations": {
        "annotation_key1": "annotation_key2"
    }
}
```

It's assumed that the annotations is a plain string:string map, with no nested structures, according to the [Helm specification](https://github.com/helm/helm/blob/657ce552cb6e582976c08cccc9605e42c242084e/pkg/chart/metadata.go#L71-L73). Setting annotations in a different format will make the Helm release fail itself (`ArtifactFailed` error), with no registered event by the Helm release controller, so there's no point on checking for these.

# Summary of changes
* Add `metadata *chart.Metadata` parameter to `event` function in `controllers/helmrelease_controller.go`
* In `reconcile` function, modify the calls to `event` in to pass a `nil` value for this new param
* In `reconcileRelease` function, modify the calls to `event` in passing `chart.Metadata`
* In `event` function, initialize empty `meta` map at the beginning
* In `event` function, iterate through the newly added `chart.Metadata` parameter if not null, and merge the data to the `meta` map that gets send as part of the event

# Testing

## Manual testing

Below are the steps I followed for testing it with a local k8s cluster. The infrastructure and source code repositores I've used are private, but they're just the minimal for configuring helm releases and alerts, and a simple API server to make test releases for, respectively. I can provide more details if needed.

The source code contains an endpoint for which the HelmRelease notifications are configured for, which logs the request payload, so that there's no need for another service for the notifications in order to check the payload.

The following versions of the tools were used:
* minikube v1.25.2
* kubectl client v1.24.2
* flux v0.31.1

### Setting the local cluster up

Start minikube:

```bash
minikube start
```

(Once the cluster is up, the source code docker image needs to be built loaded to minikube).

On the infra repository, bootstrap Flux:

```bash
flux bootstrap gitlab \
  --owner=julenpardo \
  --hostname=gitlab.hostname.com \
  --repository=fleet-fastapi-example \
  --branch=main \
  --path=./clusters/my-cluster \
  --personal
```

### Running the Helm controller

Scale the already running Helm controller down:

```bash
kubectl -n flux-system scale deployment/helm-controller --replicas=0
```

Forward the source and notification controllers:

```bash
kubectl -n flux-system port-forward svc/source-controller 8081:80
kubectl -n flux-system port-forward svc/notification-controller 8082:80
```

Run the controller specifying the event address. Apparently the controller doesn't look for an environment variable for it unlike with the source controller, so I had to figure out how to do it, and I manually tweaked the Makefile the following way for testing purposes (I'm not sure of what the best workflow for this is):

```diff
diff --git a/Makefile b/Makefile
index f4bfbf3..18d5306 100644
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ manager: generate fmt vet
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
-	go run ./main.go
+	go run ./main.go --events-addr=http://localhost:8082
 
 # Install CRDs into a cluster
 install: manifests
```
Finally, run the controller:

```bash
export SOURCE_CONTROLLER_LOCALHOST=localhost:8081 
make run
```

### Test: watch logs for Helm releases

Watch the logs of the `fastapi-example` pod:

```bash
kubectl logs -n fastapi -f $(kubectl get pods -n fastapi -o=name)
```

In the `fastapi-example` project directory, define a new Helm release by bumping `version` in `charts/Chart.yaml`. Also add some annotations, e.g.:

```yaml
annotations:
    env: dev
    commit_sha: eb93c4033dc402de9d97c3d345c6ef1bce7d125f
    repo_host: https://github.com
    repo_owner: julenpardo
    repo_name: fastapi-exapmle
```

Commit, push and wait for the Helm controller to make the upgrade. Once done, the notification endpoint should have been hit, and the `fastapi-example` log output should show the payload including the annotations within `metadata`:

```json
{
    "involvedObject": {
        "kind": "HelmRelease",
        "namespace": "flux-system",
        "name": "fastapi",
        "uid": "e92d963d-3c45-4a53-a2c5-11df902fe1f5",
        "apiVersion": "helm.toolkit.fluxcd.io/v2beta1",
        "resourceVersion": "3551"
    },
    "severity": "info",
    "timestamp": "2022-08-04T17:23:50Z",
    "message": "Helm upgrade succeeded",
    "reason": "info",
    "metadata": {
        "commit_sha": "eb93c4033dc402de9d97c3d345c6ef1bce7d125f",
        "env": "dev",
        "repo_host": "https://github.com",
        "repo_name": "fastapi-exapmle",
        "repo_owner": "julenpardo",
        "revision": "1.0.2"
    },
    "reportingController": "helm-controller",
    "reportingInstance": "jpardo-le-XXXXX"
}
```

(The `Helm upgrade has started` has been omitted for convenience, but the same metadata is included in the payload).

Make another release this time without any annotations, to check that everything is still okay without them, and check the logs:

```json
{                                                                                                                                 
    "involvedObject": {                                                                                                           
        "kind": "HelmRelease",                                                                                                    
        "namespace": "flux-system",                                                                                               
        "name": "fastapi",                                                                                                        
        "uid": "e92d963d-3c45-4a53-a2c5-11df902fe1f5", 
        "apiVersion": "helm.toolkit.fluxcd.io/v2beta1",
        "resourceVersion": "5850"
    },                 
    "severity": "info",                                          
    "timestamp": "2022-08-04T17:40:01Z",
    "message": "Helm upgrade succeeded",
    "reason": "info",
    "metadata": {          
        "revision": "1.0.3"
    },                                                           
    "reportingController": "helm-controller",
    "reportingInstance": "jpardo-le-XXXXX"
}  
```

## Unit and regression testing

I didn't add any automated tests mainly because currently there's no coverage for the flow for `reconcileRelease`, so I'm not sure of what would the best approach be to tackle this, and not end up being a pull request of adding tests rather than a feature. But suggestions are welcome.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -1,5 +1,7 @@
 /*
 Copyright 2020 The Flux authors
+Copyright 2022, DataRobot, Inc. Modified the original to include Helm chart
+annotations as event metadata.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -225,20 +227,20 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, hr v2.HelmRelease
 	if reconcileErr != nil {
 		if acl.IsAccessDenied(reconcileErr) {
 			log.Error(reconcileErr, "access denied to cross-namespace source")
-			r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, reconcileErr.Error())
+			r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, reconcileErr.Error(), nil)
 			return v2.HelmReleaseNotReady(hr, apiacl.AccessDeniedReason, reconcileErr.Error()),
 				ctrl.Result{RequeueAfter: hr.Spec.Interval.Duration}, nil
 		}
 
 		msg := fmt.Sprintf("chart reconciliation failed: %s", reconcileErr.Error())
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, msg)
+		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityError, msg, nil)
 		return v2.HelmReleaseNotReady(hr, v2.ArtifactFailedReason, msg), ctrl.Result{Requeue: true}, reconcileErr
 	}
 
 	// Check chart readiness
 	if hc.Generation != hc.Status.ObservedGeneration || !apimeta.IsStatusConditionTrue(hc.Status.Conditions, meta.ReadyCondition) {
 		msg := fmt.Sprintf("HelmChart '%s/%s' is not ready", hc.GetNamespace(), hc.GetName())
-		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityInfo, msg)
+		r.event(ctx, hr, hr.Status.LastAttemptedRevision, events.EventSeverityInfo, msg, nil)
 		log.Info(msg)
 		// Do not requeue immediately, when the artifact is created
 		// the watcher should trigger a reconciliation.
@@ -250,7 +252,7 @@ func (r *HelmReleaseReconciler) reconcile(ctx context.Context, hr v2.HelmRelease
 		if err := r.checkDependencies(hr); err != nil {
 			msg := fmt.Sprintf("dependencies do not meet ready condition (%s), retrying in %s",
 				err.Error(), r.requeueDependency.String())
-			r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityInfo, msg)
+			r.event(ctx, hr, hc.GetArtifact().Revision, events.EventSeverityInfo, msg, nil)
 			log.Info(msg)
 
 			// Exponential backoff would cause execution to be prolonged too much,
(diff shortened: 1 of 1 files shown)


stefanprodan (commented):
Thanks @julenpardo for this contribution.

I think we need to prefix all keys with `helm` to avoid collisions in notification-controller. This PR will have to wait till end of August as Hidde is in vacation and he will probably want this to be merged into the dev branch.

julenpardo:
@stefanprodan thanks a lot for such a quick response!

Should we then wait for Hidde to discuss how to prefix the keys?

hiddeco:
Thank you for your contribution @julenpardo.

I am interested in adding this to the current refactoring work which lives in `dev`, but not all events have materialized there yet (which is something I am working on in addition to extensive test coverage now that I have returned). Once my PR is up which adds the majority of the events, I will circle back to this and see how we can properly pass on and/or prefix the data.

julenpardo:
Thanks a lot @hiddeco for checking it and the heads up!

matheuscscp:
Hi @julenpardo thanks for putting this together, it's gonna be very useful!

A question: Where exactly do these chart annotations come from? Do they come from the `HelmRelease` `.spec`, `.metadata`, or `.spec.chart.spec`, or somewhere else?

I would really appreciate a way in which I could specify metadata in my `HelmRelease` manifest, like this: https://github.com/fluxcd/helm-controller/pull/682

Is your intention to also cover something like that or very similar? Could you please provide an example on where would I put my annotations/metadata exactly?

Thanks one more time!!

moritzschmitz-oviva:
> A question: Where exactly do these chart annotations come from? Do they come from the `HelmRelease` `.spec`, `.metadata`, or `.spec.chart.spec`, or somewhere else?

I believe this PR revolves around the chart annotations in the `Chart.yaml`, like [here](https://github.com/bitnami/charts/blob/main/bitnami/keycloak/Chart.yaml#L4).

stefanprodan:
Adding custom metadata to the events will be available in Flux 2.5. See [RFC-0008](https://github.com/fluxcd/flux2/tree/main/rfcs/0008-custom-event-metadata-from-annotations ) for more details.

PR #517, opened 2022-08-13
fuzz: Refactor Fuzz tests based on Go native fuzzing

pjbgf:
Moving into Go Native, the adhoc changes and on-demand build is no
longer necessary.

Previously calls to r.EventRecorder.AnnotatedEventf resulted in panic.
The new dummy recorder resolves the problem without impacting
resource consumption.

A new make target `fuzz-native` was introduced, to loop through all
fuzz tests for the duration of time specified via the environment
variable `FUZZ_TIME`.

Part of https://github.com/fluxcd/flux2/issues/2417.


pjbgf:
Added an additional commit (76fd0abd9d7a489671f412e233db1ef74fb98871) to fix issue with non-fuzz funcs.
The same fix was [proposed upstream](https://github.com/google/oss-fuzz/pull/8285), so I move this back to draft in the mean time.

hiddeco (approved):

PR #520, opened 2022-08-16
Add validation to TargetPath and ValuesKey

pjbgf:
Formalises the API requirements around TargetPath and ValuesKey,
which were the two fields missing validation within ValuesReference.
In both cases the validation was introduced at CRD level, so that
the apiserver will enforce it. 

ValuesKey must be a valid Data Key. Therefore the same logic used by
upstream Kubernetes is reused here to ensure a valid key is being used.

For TargetPath a loose regex is being used to largely represent the
expected format. A max length of 250 is now being enforced.

This is a breaking change, as invalid TargetPath and ValuesKey will now
be rejected by the apiserver, instead of being accepted and potentially
failing at reconciliation time.

### DIFF
--- a/controllers/helmrelease_controller_test.go
+++ b/controllers/helmrelease_controller_test.go
@@ -19,7 +19,9 @@ package controllers
 import (
 	"context"
 	"reflect"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/go-logr/logr"
 	"helm.sh/helm/v3/pkg/chartutil"
@@ -27,6 +29,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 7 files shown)


makkes (approved):

controllers/helmrelease_controller.go
```diff
@@ -65,6 +67,11 @@ import (
 	"github.com/fluxcd/helm-controller/internal/util"
 )
 
+var (
+	targetPathMaxLen = 250
+	targetPathRegex  = regexp.MustCompilePOSIX(`^([a-zA-Z0-9_.\\\/]|\[[0-9]{1,5}\])+$`)
```
stefanprodan, line 72: We could set these in the CRD as kubebuilder annotations, for the Kubernetes API to perform the validation instead of us.

stefanprodan:
The kubeconfig.key could use the same validation.

controllers/helmrelease_controller.go
pjbgf: Updated the PR accordingly. The only downside of this approach is that existing objects which are made invalid by the new CRD will need to be abide by the new validation before the API Server allows them to be deleted.

That being said, the new rules are largely to formalise the expected format and length, and are not expected to break existing setups.

pjbgf:
> The kubeconfig.key could use the same validation.

Let's do that as a separate PR.

makkes (commented):
Thanks a lot for clarifying the context of this change!

stefanprodan (approved):
LGTM

Thanks @pjbgf

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
 	}
 }
 
+func TestValidation(t *testing.T) {
+	tests := []struct {
+		name       string
+		resources  []runtime.Object
+		references []v2.ValuesReference
+		values     string
```
darkowlzz, line 289: It doesn't seem to be used in any of the cases below.

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
...
+		wantErr    bool
+	}{
+		{
+			name: "valid ValuesKey",
+			resources: []runtime.Object{
+				valuesSecret("values", map[string][]byte{"single": []byte("true")}),
```
darkowlzz, line 295: This value never changes, how about putting this in a variable and reusing it in all the cases. Would be easy to update in the future if we need to.

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
 	}
 }
 
+func TestValidation(t *testing.T) {
```
darkowlzz, line 284: The tests specifically validate values references. Let's rename it to be more specific.

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
...
+		{
+			name: "valid ValuesKey: long",
+			resources: []runtime.Object{
+				valuesSecret("values", map[string][]byte{"single": []byte("true")}),
+			},
+			references: []v2.ValuesReference{
+				{
+					Kind:      "Secret",
+					Name:      "values",
+					ValuesKey: func() string { return strings.Repeat("a", 253) }(),
```
darkowlzz, line 329: why not just
```suggestion
					ValuesKey: strings.Repeat("a", 253),
```

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
...
+			references: []v2.ValuesReference{
+				{
+					Kind:       "Secret",
+					Name:       "values",
+					ValuesKey:  "single",
+					TargetPath: "a[",
+				},
+			},
+			wantErr: true,
+		},
```
darkowlzz, line 447: This and the case above appear to be the same.

controllers/helmrelease_controller_test.go
```diff
@@ -278,6 +281,208 @@ invalid`,
...
+					ValuesFrom: tt.references,
+					Values:     values,
+				},
+			}
+
+			err := k8sClient.Create(context.TODO(), &hr, client.DryRunAll)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("composeValues() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
darkowlzz, line 481: How about deleting the objects at the end? I think we had some discussion recently about creating a new namespace per object and then deleting everything.

controllers/helmrelease_controller_test.go
darkowlzz: Or maybe this need not be in test cases and just be a static value.

controllers/helmrelease_controller_test.go
stefanprodan: Yes please, let's delete all resources using deferral.

pjbgf: This test only needs to validate the resources, we don't need to create them. Therefore we are using `client.DryRunAll`, so there is no objects to delete.

controllers/helmrelease_controller_test.go
pjbgf: Added a new test case.

controllers/helmrelease_controller_test.go
pjbgf: Resources are not required now that we are only validating the API level validations, and therefore has been removed.

darkowlzz (approved):
LGTM!

Issue #547, opened 2022-10-21
Interval or timeout >= 60m fails to delete


samgiles:
We just upgraded to the latest version of gotk and noted the breaking change:

> Strict validation rules have been put in place for API fields which define a time duration, such as .spec.interval. Effectively, this means values without a time unit (e.g. ms, s, m, h) will now be rejected by the API server.

However it seems:

 1. `h` values are forbidden `should match '^([0-9]+(\\.[0-9]+)?(ms|s|m))+$'`
 2. The internal representation for `m` values greater than or equal to 60 are turned into `h` representations 🤔 

Example spec snippet, it's possible to create the release, but not delete.

```yaml
...
spec:
  chart:
    spec:
      chart: my-chart
      interval: 30m
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: my-helm-charts
        namespace: my-namespace
      version: 10.1.2
  install:
    remediation:
      retries: 12
    timeout: 60m
  interval: 60m
  releaseName: my-release
  targetNamespace: my-namespace
  timeout: 60m
...
```

It seems like the timeout can't be set using `h` in the latest release, so I changed to `m` however it seems like some internal value becomes `1h0m0s` and then the controller fails. I noticed this as a the controller is unable to delete the resources although they were created successfully.

```
{"level":"error","ts":"2022-10-21T10:52:11.149Z","msg":"Reconciler error","controller":"helmrelease","controllerGroup":"helm.toolkit.fluxcd.io","controllerKind":"HelmRelease","HelmRelease":{"name":"my-release","namespace":"my-namespace"},"namespace":"my-namespace","name":"my-release","reconcileID":"cac4864b-8902-4f52-ad4f-7cc5fa540f9a","error":"HelmRelease.helm.toolkit.fluxcd.io \"my-release\" is invalid: [spec.timeout: Invalid value: \"1h0m0s\": spec.timeout in body should match '^([0-9]+(\\.[0-9]+)?(ms|s|m))+$', spec.install.timeout: Invalid value: \"1h0m0s\": spec.install.timeout in body should match '^([0-9]+(\\.[0-9]+)?(ms|s|m))+$']"}
```

I couldn't find a duplicate issue for this, apologies if this has been addressed.

samgiles:
I've raised a PR albeit unlikely to be complete. 

[...]

PR #548, opened 2022-10-21
Allow timeouts to be specified with `h` values

samgiles:
Not certain this is a complete PR as I'm unfamiliar with the codebase. Addresses https://github.com/fluxcd/helm-controller/issues/547

There's an argument that timeouts shouldn't be so large, but this validation doesn't seem like the right place to enforce that.


hiddeco:
Superseded by #549

PR #549, opened 2022-10-21
api: allow configuration of `h` unit for timeouts

### LINKED ISSUE #547: Interval or timeout >= 60m fails to delete

hiddeco:
Unlike the other Flux components, there are some scenarios in which a timeout of 1h (or more) is justified to be able to Helm actions which take a long time to finish.

To support this, widen the validation added in
54aee36b3a9adaade8625965604b5d941f63dcd3 to allow the configuration of the `h` time unit again. At the cost of theoretically keeping the workers busy for a long duration of time (and thereby not processing other changes), but without causing a full denial of service.

Follow up on #533
Fixes #547 https://github.com/fluxcd/flux2/issues/3227

### DIFF
--- a/api/v2beta1/helmrelease_types.go
+++ b/api/v2beta1/helmrelease_types.go
@@ -333,7 +333,7 @@ type Install struct {
 	// Jobs for hooks) during the performance of a Helm install action. Defaults to
 	// 'HelmReleaseSpec.Timeout'.
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m))+$"
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m|h))+$"
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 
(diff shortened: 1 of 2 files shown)


stefanprodan:
We need to do the same in KC, the health checks would take as long as the Helm wait.

stefanprodan (approved):

PR #584, opened 2022-12-21
Align graceful-shutdown-timeout with terminationGracePeriodSeconds

relu:
Setting the default value for the graceful-shutdown-timeout flag to match the default terminationGracePeriodSeconds value we set for the controller pod container.
It seems the controller-runtime does not support passing -1 as a value to skip the timeout as documented here:
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.13.1/pkg/manager/manager.go#L286

With this change, it would also make sense to update the documentation to add a new section about graceful shutdown and explain the relation between these two values and also related to how the readiness probe affects graceful shutdown.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -48,7 +48,7 @@ jobs:
         uses: engineerd/setup-kind@v0.5.0
         with:
           version: v0.11.1
-          image: kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
Thanks for getting to the bottom of this @relu :bow:

stefanprodan:
@relu to fix the e2e tests please do this https://github.com/fluxcd/flux2/commit/21a943e6f9db133cbe3023c21b5ee071ea56c460

stefanprodan (approved):

stefanprodan:
We also need to update the options.md in website.

PR #596, opened 2023-01-31
Patch CRDs with origin labels

### LINKED ISSUE #476: CRDs are not labeled
When a HelmRelease contains CRDs, they get applied on the cluster without the owner labels `helm.toolkit.fluxcd.io/name` and `helm.toolkit.fluxcd.io/namespace`.

hiddeco:
This allows the applied CRDs to be traced using the same labels as currently applied to resources using a Kustomize post-render.

Kustomize is not used here as the apply logic for CRDs is different from the approach used during releasing, where we inject the labels in such a way that they are written back to the Helm storage in the rendered manifest. This to match Helm's logic from which our present code is already derived (buth with support for policies).

This also moves the full responsibility of dealing with the install of CRDs to ourselves, as we no longer fall back to Helm's logic when `Create` is configured as a policy during a Helm install. As this would not allow us to add the labels.

Fixes #476

### DIFF
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -38,11 +38,14 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

Issue #186, opened 2021-01-07
Cluster-state manifest drift correction


ruchikaguptaa:
Suppose `HelmRelease` object deploys a helm chart that contains a deployment. The pod resource can now be updated by the user. Until there is a chart version update in `HelmRelease` object, the pod will continue to have the manually updated setting. I'm looking for a way to force reconcile the resources deployed by the `HelmRelease` after every x minutes. Does helm-controller support this?

There are a few other intervals like one used by HelmRepository and HelmRelease but they don't seem to reconcile a previous succeeded release unless there is a new version in the HelmRelease. `requeueDependency` interval is only responsible for reconcile if the dependencies for that helmrelease failed. These seem to be unrelated to my scenario.

ruchikaguptaa:
Tagging a few folks as per the PRs to the reconcile helmchart part of code @hiddeco @seaneagan .

hiddeco:
The helm-controller does not support this at present, nor did the Helm Operator.

I have been playing around with proof of concepts to determine changes based on the manifest blob that is stored in the Helm storage based on 2 and 3-way merges, but ran into some issues around mutating cluster webhooks that may modify the object when it is applied on the cluster by Helm.

brondum:
@hiddeco any updates, just realised that a deployment was deleted but helm was yay happy.
Could be super cool if flux could detect the state was undesired :)

pedroparraortega:
we are looking forward to have this feature!! Any updates?

cmd-werner:
Facing the same issue. Reconciliation does not work with resources deployed by Helm.

rxbn:
i'm also looking forward to have this feature! what's the current status?

tmk1991:
You could try to do a cronjob with manual reconciliation 

`flux reconcile hr {{your hr}} --with-source -n {{your ns}}`

Jeinhaus:
The only way to restore the helmrelease to the defined state is to reinstall or upgrade it, right?

Legion2:
I deleted a crd (and all its objects) and recreated it, however helm controller does not recreate the objects from the helm release. Not my cluster is in an inconsistent state and I does not recover. Even if I call `flux reconcile helmrelease` it does not recreate the deleted resources from the helm template.

hiddeco:
[...]

PR #617, opened 2023-02-20
Experimental support of drift detection

hiddeco:
This enables experimental drift detection of cluster state compared to
the current manifest data from the Helm storage's manifest blob
by starting the controller with `--feature-gates=DetectDrift=true`.

Drift detection works based on the already proven approach of the
kustomize-controller's SSA package, and utilizes the managed field
configured by the controller since `v0.12.2`.

This feature is planned to go out of experimental once the further
controller rewrite has been finished, and the state of the Helm storage
itself is more fault tolerant.

It allows a specific object from a release manifest to be excluded
from drift detection by label or annotating it with:
`helm.toolkit.fluxcd.io/diff: disabled`.

Using a Kustomize post renderer definition in a HelmRelease, this can
be used to ignore any object from an arbitrary chart.

Experimentally covers #186


docs/spec/v2beta1/helmreleases.md
```diff
@@ -1270,6 +1270,42 @@ spec:
...
+When a HelmRelease is in-sync with the Helm release object in the storage, the controller will
+compare the manifests from the Helm storage with the current state of the cluster using a
+[server-side dry-run apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/).
+If this comparison detects a drift (either due resource being created or modified during the
+dry-run), the controller will perform an upgrade for the release, restoring the desired state.
+
+### Excluding resources from drift detection
+
+The drift detection feature can be configured to exclude certain resources from the comparison
+by labeling or annotating them with `helm.toolkit.fluxcd.io/diff: disabled`. Using
```
hiddeco, line 1286: Open to suggestions to a better exclusion label, know that @stefanprodan wants to prevent "true" / "false" values and prefers something that matches with "enabled" / "disabled".

internal/diff/differ.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		switch entry.Action {
+		case ssa.CreatedAction, ssa.ConfiguredAction:
+			diff = true
+			changeSet.Add(*entry)
+
+			if entry.Action == ssa.ConfiguredAction {
+				// TODO: remove this once we have a better way to log the diff
+				//       for example using a custom dyff reporter, or a flux CLI command
```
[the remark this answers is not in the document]
hiddeco: I really meant [dyff](https://github.com/homeport/dyff) here, which is also what is being used for `flux diff`.

docs/spec/v2beta1/helmreleases.md
hiddeco: Annotations are `camelCase`, but you're on the same mental journey as me.

internal/diff/differ.go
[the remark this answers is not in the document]
aryan9600: oh okay TIL

docs/spec/v2beta1/helmreleases.md
pjbgf: +1 on `helm.toolkit.fluxcd.io/drift-detection`.

hiddeco: This would then become `helm.toolkit.fluxcd.io/driftDetection`, as we also have `reconcile.fluxcd.io/requestedAt`.

internal/diff/differ.go
```diff
@@ -0,0 +1,155 @@
...
+		entry, releaseObject, clusterObject, err := resourceManager.Diff(ctx, obj, ssa.DiffOptions{
+			Exclusions: map[string]string{
+				MetadataKey: MetadataDisabledValue,
+			},
+		})
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		switch entry.Action {
```
pjbgf, line 123: entry could be `nil`, may want to add a check above.

internal/cmp/simple_unstructured.go
```diff
@@ -0,0 +1,110 @@
...
+func pathToString(path cmp.Path) (p string) {
+	for _, step := range path {
+		switch t := step.(type) {
+		case cmp.MapIndex:
+			p += fmt.Sprintf(".%v", t.Key())
+		case cmp.SliceIndex:
+			p += fmt.Sprintf("%v", t.String())
+		}
+	}
+	p = strings.TrimPrefix(p, ".")
```
pjbgf, line 108: completely understand the extraction of `pathToString` to a standalone `func`, just wondering whether there would be a nice solution to reuse the existing `StringBuilder` instance here.

pjbgf (approved):
Awesome job @hiddeco!

internal/cmp/simple_unstructured.go
hiddeco: I could theoretically change it to something like:

```go
func writePathString(path cmp.Path, sb strings.Builder) {
	for _, step := range path {
		switch t := step.(type) {
		case cmp.MapIndex:
			sb.WriteString(fmt.Sprintf(".%v", t.Key()))
		case cmp.SliceIndex:
			sb.WriteString(fmt.Sprintf("%v", t.String()))
		}
	}
	return
}
```

At the cost of paths starting with a `.` (practically full YAML dot notation), but this may actually be better to the eyes when reading a lengthy diff output.

stefanprodan (approved):
LGTM

Thanks @hiddeco 🥇 

PS. We need to add a section to the bootstrap docs and show how users can enable this feature.

PR #623, opened 2023-03-01
check if service account exists before uninstalling release

aryan9600:
Check if the service account to be impersonated actually exists
and proceed with uninstalling the Helm release only if it does.
Otherwise, skip uninstalling the release and carry on with finalization.
Add an e2e test to check if deleting a namespace with the RBAC and
HelmRelease succeeds with the namespace being fully deleted.

### DIFF
--- a/controllers/helmrelease_controller.go
+++ b/controllers/helmrelease_controller.go
@@ -660,8 +660,11 @@ func (r *HelmReleaseReconciler) composeValues(ctx context.Context, hr v2.HelmRel
 
 // reconcileDelete deletes the v1beta2.HelmChart of the v2beta1.HelmRelease,
 // and uninstalls the Helm release if the resource has not been suspended.
+// It only performs a Helm uninstall if the ServiceAccount to be impersonated
+// exists.
 func (r *HelmReleaseReconciler) reconcileDelete(ctx context.Context, hr v2.HelmRelease) (ctrl.Result, error) {
 	r.recordReadiness(ctx, hr)
+	log := ctrl.LoggerFrom(ctx)
 
 	// Delete the HelmChart that belongs to this resource.
 	if err := r.deleteHelmChart(ctx, &hr); err != nil {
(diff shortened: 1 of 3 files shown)


controllers/helmrelease_controller.go
```diff
@@ -678,11 +683,42 @@ func (r *HelmReleaseReconciler) reconcileDelete(ctx context.Context, hr v2.HelmR
...
+			hr.Spec.ServiceAccountName,
+			hr.GetNamespace(),
+		)
+		if impersonator.CanImpersonate(ctx) {
+			if err := run.Uninstall(hr); err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
+				return ctrl.Result{}, err
+			}
+			log.Info("uninstalled Helm release for deleted resource")
+		} else {
+			release, err := run.ObserveLastRelease(hr)
```
hiddeco, line 703: Did you test this? I would expect this to not be able to reach the Secret, as the `getter` configured makes use of impersonation as well.

aryan9600: yes i have and it works fine. the getter does make use of the impersonation, but it doesn't actually need the ServiceAccount to exist as long as there exists a RoleBinding which points to a Role with sufficient perms.

stefanprodan: To test this, create a HR in a namespace along with the SA and the role binding, delete the namespace and see it stuck in termination.

stefanprodan (commented):
Please add an e2e test for namespace deletion, where we check that the namespace has been deleted, thus proving that we remove the finalizer.

controllers/helmrelease_controller.go
```diff
@@ -670,19 +673,35 @@ func (r *HelmReleaseReconciler) reconcileDelete(ctx context.Context, hr v2.HelmR
...
+				return ctrl.Result{}, err
+			}
+			if err := run.Uninstall(hr); err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
+				return ctrl.Result{}, err
+			}
+			log.Info("uninstalled Helm release for deleted resource")
+		} else {
+			msg := "skipping Helm uninstall"
+			log.Error(fmt.Errorf("failed to find service account to impersonate"), msg)
+			r.event(ctx, hr, hr.Status.LastAppliedRevision, eventv1.EventSeverityError, msg)
```
hiddeco, line 703:
```suggestion
			err := fmt.Errorf("failed to find service account to impersonate")
			msg := "skipping Helm uninstall"
			log.Error(err, msg)
			r.event(ctx, hr, hr.Status.LastAppliedRevision, eventv1.EventSeverityError, fmt.Sprintf("%s: %s", msg, err.Error()))
```

stefanprodan (approved):
LGTM

Thanks @aryan9600 🥇

hiddeco (approved):
Thank you @aryan9600 :bow:

PR #625, opened 2023-03-01
runner: configure Helm action cfg log levels

hiddeco:
This reduces the amount of log lines pushed to `debug` by configuring the kube client and storage loggers to only log to `trace`.

In addition, the log buffer used in events will now just contain the most relevant information about a failure as reported by the Helm action itself, and not the in-depth information from the underlying client and/or storage.

### DIFF
--- a/internal/runner/log_buffer.go
+++ b/internal/runner/log_buffer.go
@@ -28,16 +28,10 @@ import (
 
 const defaultBufferSize = 5
 
-type DebugLog struct {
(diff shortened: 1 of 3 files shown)


hiddeco:
@stefanprodan if you quickly want to test this, you can use `docker.io/hiddeco/helm-controller:log-levels`.

stefanprodan (approved):

Issue #149, opened 2020-11-19
Helm upgrade failed: another operation (install/upgrade/rollback) is in progress


runningman84:
Sometimes helm releases are not installed because of this error:

```
{"level":"info","ts":"2020-11-19T15:41:11.273Z","logger":"controllers.HelmRelease","msg":"reconcilation finished in 50.12655ms, next run in 9m0s","controller":"helmrelease","request":"traefik/traefik"}
{"level":"error","ts":"2020-11-19T15:41:11.274Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"helm.toolkit.fluxcd.io","reconcilerKind":"HelmRelease","controller":"helmrelease","name":"traefik","namespace":"traefik","error":"Helm upgrade failed: another operation (install/upgrade/rollback) is in progress"}
{"level":"info","ts":"2020-11-19T15:43:19.310Z","logger":"controllers.HelmRelease","msg":"reconcilation finished in 69.439664ms, next run in 9m0s","controller":"helmrelease","request":"traefik/traefik"}
{"level":"error","ts":"2020-11-19T15:43:19.310Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"helm.toolkit.fluxcd.io","reconcilerKind":"HelmRelease","controller":"helmrelease","name":"traefik","namespace":"traefik","error":"Helm upgrade failed: another operation (install/upgrade/rollback) is in progress"}
{"level":"info","ts":"2020-11-19T15:52:42.524Z","logger":"controllers.HelmRelease","msg":"reconcilation finished in 69.944579ms, next run in 9m0s","controller":"helmrelease","request":"traefik/traefik"}
{"level":"error","ts":"2020-11-19T15:52:42.525Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"helm.toolkit.fluxcd.io","reconcilerKind":"HelmRelease","controller":"helmrelease","name":"traefik","namespace":"traefik","error":"Helm upgrade failed: another operation (install/upgrade/rollback) is in progress"}
```

In this case the helm release is stuck in pending status.

We have not found any corresponding log entry of the actual installation. Is this some concurrency bug?

hiddeco:
Is this happening so often that it would be possible to enable the `--log-level=debug` flag for awhile so we get better insight into what Helm exactly does?

seaneagan:
[...]

PR #628, opened 2023-03-03
Introduce OOM watcher to allow graceful shutdown

hiddeco:
This PR introduces an OOM watcher, which can be enabled using
`--feature-gates=OOMWatch=true`. The OOM watcher watches the current
memory usage as reported by cgroups via `memory.current` and cancels
the context when it reaches a certain threshold compared to
`memory.max` (default `95`%, configurable using
`--oom-watch-memory-threshold`).

This allows ongoing Helm processes to gracefully exit with a failure
before the controller is forcefully OOM killed, preventing a deadlock
of releases in a pending state.

The OOM watcher polls the `memory.current` file on an interval (default
`500ms`, configurable using `--oom-watch-interval`), as subscribing to
file updates using inotify is not possible for cgroups (v2) except for
`*.events` files. Which does provide signals using `memory.events`, but
these will generally be too late for our use case as for example `high`
equals `max` in most containers, buying us little time to gracefully
stop our processes.

In addition, because we simply watch current usage compared to max
usage in bytes. This approach should work for cgroups v1 as well, given
this has (most of the time) files for these values available as well,
albeit at times at different locations (for which this commit does not
introduce a flag yet, but the library takes into account that it could
be configured at some point).

Should help aid #149 when enabled.

### DIFF
--- a/internal/oomwatch/watch.go
+++ b/internal/oomwatch/watch.go
@@ -47,7 +47,7 @@ type Watcher struct {
 	memoryCurrentPath string
 	// memoryUsagePercentThreshold is the threshold at which the system is
 	// considered to be near OOM.
-	memoryUsagePercentThreshold float64
+	memoryUsagePercentThreshold uint8
 	// interval is the interval at which to check for OOM.
 	interval time.Duration
 	// logger is the logger to use.
@@ -62,9 +62,13 @@ type Watcher struct {
 }
 
 // New returns a new Watcher.
-func New(memoryMaxPath, memoryCurrentPath string, memoryUsagePercentThreshold float64, interval time.Duration, logger logr.Logger) (*Watcher, error) {
+func New(memoryMaxPath, memoryCurrentPath string, memoryUsagePercentThreshold uint8, interval time.Duration, logger logr.Logger) (*Watcher, error) {
 	if memoryUsagePercentThreshold < 1 || memoryUsagePercentThreshold > 100 {
-		return nil, fmt.Errorf("memory usage percent threshold must be between 1 and 100, got %.2f", memoryUsagePercentThreshold)
+		return nil, fmt.Errorf("memory usage percent threshold must be between 1 and 100, got %d", memoryUsagePercentThreshold)
+	}
+
+	if minInterval := 50 * time.Millisecond; interval < minInterval {
+		return nil, fmt.Errorf("interval must be at least %s, got %s", minInterval, interval)
 	}
 
 	if _, err := os.Lstat(memoryCurrentPath); err != nil {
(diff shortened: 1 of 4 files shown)


internal/oomwatch/watch.go
```diff
@@ -0,0 +1,168 @@
...
+	// ctx is the context that is canceled when OOM is detected.
+	ctx context.Context
+	// cancel is the function that cancels the context.
+	cancel context.CancelFunc
+	// once is used to ensure that Watch is only called once.
+	once sync.Once
+}
+
+// New returns a new Watcher.
+func New(memoryMaxPath, memoryCurrentPath string, memoryUsagePercentThreshold float64, interval time.Duration, logger logr.Logger) (*Watcher, error) {
```
pjbgf, line 65: I wonder whether `memoryUsagePercentThreshold` needs to be as granular to hold fractions of a percent. Maybe`uint8` suffices here.

internal/oomwatch/watch.go
```diff
@@ -0,0 +1,168 @@
...
+	}, nil
+}
+
+// NewDefault returns a new Watcher with default path values.
+func NewDefault(memoryUsagePercentThreshold float64, interval time.Duration, logger logr.Logger) (*Watcher, error) {
+	return New(
+		filepath.Join(DefaultCgroupPath, MemoryMaxFile),
+		filepath.Join(DefaultCgroupPath, MemoryCurrentFile),
+		memoryUsagePercentThreshold,
+		interval,
```
pjbgf, line 94: Thoughts on maybe providing a sensible default for `interval` instead of requiring one through this route.

internal/oomwatch/watch.go
```diff
@@ -0,0 +1,168 @@
...
+	memoryMax, err := readUintFromFile(memoryMaxPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read memory.max %q: %w", memoryMaxPath, err)
+	}
+
+	return &Watcher{
+		memoryMax:                   memoryMax,
+		memoryCurrentPath:           memoryCurrentPath,
+		memoryUsagePercentThreshold: memoryUsagePercentThreshold,
+		interval:                    interval,
```
pjbgf, line 83: It may be worth validating interval, if it is too low CPU/IO would be affected. Potentially 10-50ms as min?

pjbgf (commented):
A few nits, but otherwise really good stuff. :clap:

internal/oomwatch/watch.go
hiddeco: Would not be really great as we allow it to be configured via a flag, while the paths at present are fixed.

internal/oomwatch/watch.go
hiddeco: During my testing I observed many of the same value reads at `100ms` already. Now validating to ensure it does not go lower than this.

pjbgf (approved):

stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

Issue #645, opened 2023-03-14
Report-only drift detection


tun0:
While going through https://fluxcd.io/flux/components/helm/helmreleases/#drift-detection, my conclusion is that enabling drift-detection (or rather: drift-remediation) can be a tricky thing to properly configure up-front. You'd need dig through all Helm charts in detail to see if there's any potential stuff that could be broken by enabling drift-"detection". Especially when enabling it on existing resources.

My proposal would be offer a step in between: detection of drift without trying to remediate. Based on the logs/events it'd generate, one could fine-tune `spec.postRenderers` before actually enabling the remediation of drift as well.

PR #647, opened 2023-03-17
Allow opt-out of drift correction

hiddeco:
This enhances the drift detection to allow detection without acting on it by adding a `CorrectDrift` feature gate. When set to `false` (i.e. `--feature-gates=DetectDrift=true,CorrectDrift=false`), the controller will only log the difference it detected without acting on it.

This allows for a smoother transition to drift detection, as issues with a number of existing releases can be solved before the feature is enabled in full.

Fixes: #645


hiddeco:
@darkowlzz yes, this was still on my list but was nearing my EOD on Friday.

stefanprodan (approved):

PR #658, opened 2023-03-29
Add reconciler sharding capability based on label selector

hiddeco:
With this enhancement, the controller can be configured with `--watch-label-selector`, after which only objects with this label will be reconciled by the controller.

This allows for horizontal scaling of the helm-controller, where each controller can be deployed multiple times with a unique label selector which is used as the sharding key.

Note that if you want to ensure a `HelmChart` gets created for a specific source-controller instance, you have to provide the labels for this controller in `.spec.chart.metadata.labels` of the `HelmRelease`.

Part of https://github.com/fluxcd/flux2/issues/3731


main.go
```diff
@@ -155,6 +162,11 @@ func main() {
 		disableCacheFor = append(disableCacheFor, &corev1.Secret{}, &corev1.ConfigMap{})
 	}
 
+	leaderElectionId := fmt.Sprintf("%s-%s", controllerName, "leader-election")
+	if watchOptions.LabelSelector != "" {
+		leaderElectionId = leaderelection.GenerateID(leaderElectionId, watchOptions.LabelSelector)
```
stefanprodan, line 167: This should be used at line 185

stefanprodan (approved):
LGTM

PR #662, opened 2023-03-31
config/*: update API versions and file names 

hiddeco:


stefanprodan:
Please bump podinfo everywhere to 6.3.5 (I did this for KC), this way we avoid the HPA v1beta2 deprecation errors in the future.

Issue #675, opened 2023-04-20
Flux reconcile make HR upgrade without any update


longquan0104:
My helm-release was upgraded and re-rendered to generate new resources without any update. 
My helm-release has those value
```
      test:
              632bd80235a05f4192aefade: TEST_SECRET_VALUE
              632bd80ddf416cf32fd50679: TEST_SECRET_VALUE
              632bd817c559818a52307da2: TEST_SECRET_VALUE
              632bd82398e71231a98004b6: TEST_SECRET_VALUE 
```

every time re-render it makes a new manifest with only one update position of this map key: value
The update is only about the position:
for example changes from revision 1 to 2: 
 ```
         test:
+          632bd80235a05f4192aefade: TEST_SECRET_VALUE
            632bd80ddf416cf32fd50679: TEST_SECRET_VALUE
            632bd817c559818a52307da2: TEST_SECRET_VALUE
-          632bd80235a05f4192aefade: TEST_SECRET_VALUE
            632bd82398e71231a98004b6: TEST_SECRET_VALUE
```

and changes from revision 2 to 3: 
```
        test:
-           632bd80235a05f4192aefade: TEST_SECRET_VALUE
            632bd80ddf416cf32fd50679: TEST_SECRET_VALUE
            632bd817c559818a52307da2: TEST_SECRET_VALUE
+           632bd80235a05f4192aefade: TEST_SECRET_VALUE
            632bd82398e71231a98004b6: TEST_SECRET_VALUE
```

My expectation is this should not consider an update and there should be no upgrading action happened.

PR #676, opened 2023-04-20
Fix bug checksum chart values will not consider about order of key-value

longquan0104:
Fixing this issue https://github.com/fluxcd/helm-controller/issues/675
Helm upgrade always happens without any changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,8 +85,8 @@ api-docs: gen-crd-api-reference-docs
 
 # Run go mod tidy
 tidy:
-	cd api; rm -f go.sum; go mod tidy -compat=1.20
-	rm -f go.sum; go mod tidy -compat=1.20
+	cd api; rm -f go.sum; go mod tidy -compat=1.19
+	rm -f go.sum; go mod tidy -compat=1.19
 
 # Run go fmt against code
 fmt:
--- a/Makefile
+++ b/Makefile
@@ -85,8 +85,8 @@ api-docs: gen-crd-api-reference-docs
 
 # Run go mod tidy
 tidy:
-	cd api; rm -f go.sum; go mod tidy -compat=1.19
-	rm -f go.sum; go mod tidy -compat=1.19
+	cd api; rm -f go.sum; go mod tidy -compat=1.20
+	rm -f go.sum; go mod tidy -compat=1.20
 
 # Run go fmt against code
 fmt:
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -30,6 +30,16 @@ import (
 // ValuesChecksum calculates and returns the SHA1 checksum for the
 // given chartutil.Values.
 func ValuesChecksum(values chartutil.Values) string {
+	var s string
+	if len(values) != 0 {
+		s, _ = values.YAML()
+	}
+	return fmt.Sprintf("%x", sha1.Sum([]byte(s)))
+}
+
+// OrderedValuesChecksum sort the chartutil.Values then calculates
+// and returns the SHA1 checksum for the sorted values.
+func OrderedValuesChecksum(values chartutil.Values) string {
 	var s []byte
 	if len(values) != 0 {
 		// Check sum on the formatted values
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -19,19 +19,96 @@ package util
 import (
 	"crypto/sha1"
 	"fmt"
+	"sort"
 
-	"helm.sh/helm/v3/pkg/chartutil"
+	goyaml "gopkg.in/yaml.v2"
 	"helm.sh/helm/v3/pkg/release"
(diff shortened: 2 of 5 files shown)


longquan0104:
Hello @hiddeco,
Please review this. Thank you.

Makefile
```diff
@@ -85,8 +85,8 @@ api-docs: gen-crd-api-reference-docs
 
 # Run go mod tidy
 tidy:
-	cd api; rm -f go.sum; go mod tidy -compat=1.20
-	rm -f go.sum; go mod tidy -compat=1.20
+	cd api; rm -f go.sum; go mod tidy -compat=1.19
+	rm -f go.sum; go mod tidy -compat=1.19
```
hiddeco, line 89: This should not be changed.

internal/util/util.go
```diff
@@ -19,19 +19,96 @@ package util
...
+func cleanUpMapValue(v interface{}) interface{} {
+	switch v := v.(type) {
+	case []interface{}:
+		return cleanUpInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return cleanUpInterfaceMap(v)
+	case string:
+		return v
+	default:
+		return fmt.Sprintf("%v", v)
```
hiddeco, line 82: The result of this would be that everything becomes a string type. This may have other undesirable effects, e.g. the change of an integer type to a string (or vice-versa) causing an upgrade to not be triggered.

internal/util/util.go
```diff
@@ -19,19 +19,96 @@ package util
...
+		for i, value := range newValues {
+			newValues[i] = cleanUpMapValue(value)
+		}
+		msValues := yaml.JSONObjectToYAMLObject(newValues)
+		// Sort
+		SortMapSlice(msValues)
+		// Marshal
+		s, err = goyaml.Marshal(msValues)
+		if err != nil {
+			panic(err)
```
hiddeco, line 49: Panics based on arbitrary input with little guarantee it may only occur due to programmatic errors (and not based on user-input) make me feel uncomfortable.

longquan0104:
Hi @hiddeco ,
Thank you for your comments. I updated them and please let me know if there are any issues with this. 
Thank you.

internal/util/util.go
```diff
@@ -19,19 +19,87 @@ package util
 import (
 	"crypto/sha1"
 	"fmt"
+	"sort"
 
+	goyaml "gopkg.in/yaml.v2"
```
stefanprodan, line 24: Any reason for not using v3?

hiddeco:
Have been thinking some more about this, and while I do think that it would theoretically be better to order keys to prevent unnecessary upgrades. The current way this is structured would cause any existing `HelmRelease` to upgrade when the values are ordered, even if there hasn't been a change.

Given this, I think this needs changing in such a way that this doesn't happen. For example, by calculating the checksum without ordering first, and then only calculate the version _with_ ordering if this causes a mismatch (or when an upgrade is triggered, to then take note of the ordered value).

internal/util/util.go
longquan0104: I used the v2 which is being used in _"sigs.k8s.io/yaml"_ and in the v2, they support _MapSlice_ structure which is easy to handle to me.

longquan0104:
So you mean that the key still should be ordered but there should not break the running resources and force them to upgrade?

hiddeco:
1. Calculate checksum of ordered and unordered values.
2. If the checksum as recorded in the Status matches one of those values, there isn't a change.
3. If there is a change, take note of the ordered values checksum.

This would allow us to eventually stop calculating the unordered checksum, but without triggering an upgrade right now (after upgrading to patched controller with this change).

longquan0104:
I updated the code, please let me know if there are anything need to be done @hiddeco

hiddeco:
You do not seem to have allowed for us maintainers to push changes to your branch, but please see https://github.com/fluxcd/helm-controller/commit/a33273536b92935a8113765c9962b89222c2af7d for further review reference.

internal/util/util.go
```diff
@@ -34,6 +37,81 @@ func ValuesChecksum(values chartutil.Values) string {
...
+		return cleanUpInterfaceMap(v)
+	default:
+		return v
+	}
+}
+
+func cleanUpInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	result := make(map[string]interface{})
+	for k, v := range in {
+		result[fmt.Sprintf("%T.%v", k, k)] = cleanUpMapValue(v)
```
hiddeco, line 88: Why are we taking note of the type here? It should take the checksum of an ordered map, without making any modifications to the map itself besides ordering.

longquan0104:
Hello @hiddeco, I migrate this to PR #684 which the maintainer can make commits to

Issue #680, opened 2023-04-30
Feature request: Add field to `HelmRelease` CRD for adding metadata to events


aristapimenta:
Pretty much the same ask made here: https://github.com/fluxcd/notification-controller/issues/511

Add a field to the `HelmRelease` spec for attaching more metadata to events delivered to notification-controller.

PR #682, opened 2023-05-01
Add event metadata field to HelmRelease spec

matheuscscp:
Fixes https://github.com/fluxcd/helm-controller/issues/680


hiddeco:
This has overlap with #514, and is in general quite the change to accept without more discussion among maintainers.

I suggest you join one of our [project meetings](https://fluxcd.io/community/#meetings) to further discuss your proposal.

stefanprodan:
Closing this in favour of https://github.com/fluxcd/notification-controller/pull/519

moritzschmitz-oviva:
Can we re-open this one? Yes, there is an overlap with #514, but this one here tackles a different issue. 

My use-case:
I have an app which creates [deployments](https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#about-deployments) via the GitHub API. Then in the same step, it writes a `HelmRelease` manifest to a repository which is reconciled via FluxCD. After the reconciliation, I want to use the `notification-controller` to call my app again, which then calls the deployments API again to show the deployment as done.

To be able to match the created `HelmRelease` with the deployment though, I would want to add some identifying data to the `HelmRelease` manifest and then return this as metadata via the `notification-controller`.

moritzschmitz-oviva:
@matheuscscp Any chance you joined a meeting and brought this up already?

moritzschmitz-oviva:
@matheuscscp And more on-topic, would it be enough to just include the manifest's `.metadata.labels` and/or `.metadata.annotations`?

matheuscscp:
> @matheuscscp Any chance you joined a meeting and brought this up already?

I can do it this week 👌

matheuscscp:
@moritzschmitz-oviva One question: how would you identify the GitHub deployment?

moritzschmitz-oviva:
> @moritzschmitz-oviva One question: how would you identify the GitHub deployment?

The API returns an ID when creating a deployment: https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#create-a-deployment.

We have a GitHub app which creates a deployment via above API and then modifies files inside the FluxCD repository. When we create the files, we already have the ID and can set it as a metadata field. Now when the notification-controller reports the successful installation of the HelmRelease we call the GitHub app again and pass the metadata. It then knows which deployment to 'proceed'. Does that make sense?

matheuscscp:
Hey @moritzschmitz-oviva I just wanted to let you know that we are gonna support sending object annotation keys prefixed with a to-be-defined API Group, e.g. something like `events.toolkit.fluxcd.io/myCustomMetadataKey`. This will be supported across all Flux controllers in a uniform way, not just for `HelmRelease`. I will open an RFC in the flux2 repo for this.

matheuscscp:
@moritzschmitz-oviva Feel free to comment: https://github.com/fluxcd/flux2/pull/4809/files

PR #690, opened 2023-05-24
Update Kubernetes, controller-runtime and Helm

hiddeco:
This PR updates Kubernetes to v1.27, controller-runtime to
v0.15, and Helm to v3.12.

It deals with various breaking changes in controller-runtime, as
documented in the release notes:
https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

In short:

- `Watches` now use a `client.Object` instead of a `source.Kind`.
- `handler.MapFunc` signature accepts a Go context, which is used to
  log any errors, instead of silently ignoring them and/or panicking.
- Max concurrent reconciles is configured on the manager, instead of
  configuring them per reconciler instance.
- Various manager configuration options have been moved to new
  structures and/or fields.

In addition to this, all other dependencies which had updates
available are updated to their latest versions as well.

Part of: https://github.com/fluxcd/flux2/issues/3564 and https://github.com/fluxcd/flux2/issues/3907


stefanprodan (approved):
LGTM

PR #696, opened 2023-06-01
Bump source-controller to v1.0.0-rc.5

makkes:


hiddeco (commented):
This also requires a bump in the YAML to pull in the manifests for that version, see: https://github.com/fluxcd/helm-controller/commit/564a188a7feab0e40267a8c57631af76a4c5451e

hiddeco (approved):

PR #698, opened 2023-06-09
Set deletion propagation for helm uninstall

hoffimar:
In helm 3.12 a feature was added ([commit](https://github.com/helm/helm/commit/830d4a9ee9d1655308b1a1556fd5513bedbb1731)) to support other deletion modes than background during a helm uninstall. This PR attempts to pass the new property DeletionPropagation on the HelmRelease to the runner.

* Add an optional field DeletionPropagation to the HelmRelease api.
* Set the DeletionPropagation when running the uninstall. 

I'm not so sure about the needed default handling for this optional field. To not disturb existing installations, I made sure a value is returned in case there is no default yet persisted on the HelmRelease.


hiddeco (commented):
Can you please document this new field in `docs/spec/v2beta1/helmreleases.md`?

aryan9600 (approved):
lgtm!

hiddeco (approved):
Thank you very much! 🥇

PR #703, opened 2023-06-20
Fix HelmRelease reconciliation loop

relu:
Likely after the upgrade to controller-runtime v0.15.0 a regression surfaced for long-running reconciliations of HelmRelease resources (e.g. for charts having pre-upgrade hooks taking a few minutes to complete). This regression would cause the controller to immediately re-run the upgrade after a successful upgrade, thus entering an almost-endless loop.

Apparently, the only fix to this issue is to ensure `.Status.LastReleaseRevision` is updated as soon as possible in the reconiliation cycle rather than wait for the update at the end of the cycle.


hiddeco (approved):
As I am working on rewriting this as a whole, getting this in now without deep investigation is the easiest way to go.

PR #715, opened 2023-06-26
Align `go.mod` version with Kubernetes (Go 1.20)

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4014


makkes (approved):
Might want to change the [1.18 reference](https://github.com/fluxcd/helm-controller/blob/60090cc288e62e40ed8b7bc4ad8c7a4ce9727aa6/DEVELOPMENT.md#L27) in the dev docs as well.

PR #751, opened 2023-08-09
controller: jitter requeue interval

hiddeco:
This adds a `--interval-jitter-percentage` flag to the controller to
add a +/- percentage jitter to the interval defined in a HelmRelease
(defaults to 5%).

Effectively, this results in a reconciliation every 9.5 - 10.5 minutes
for a resource with an interval of 10 minutes.

Main reason to add this change is to mitigate spikes in memory and
CPU usage caused by many resources being configured with the same
interval.

Part of: https://github.com/fluxcd/flux2/issues/4120

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,6 +41,7 @@ import (
 	helper "github.com/fluxcd/pkg/runtime/controller"
 	"github.com/fluxcd/pkg/runtime/events"
 	feathelper "github.com/fluxcd/pkg/runtime/features"
+	"github.com/fluxcd/pkg/runtime/jitter"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -89,6 +90,7 @@ func main() {
 		leaderElectionOptions     leaderelection.Options
 		rateLimiterOptions        helper.RateLimiterOptions
 		watchOptions              helper.WatchOptions
+		intervalJitterOptions     jitter.IntervalOptions
```
stefanprodan, line 93: `intervalJitterOptions.BindFlags(flag.CommandLine)` is missing

stefanprodan (approved):

Issue #772, opened 2023-09-16
Drift Detector Diff client doesn't use the defaultServiceAccount command line flag value ?


ravilr:
helm-controller version: v0.36.0

helm-controller deployed with `DetectDrift` feature gate enabled and also `--default-service-account` args like below:

```
 - args:
        - --default-service-account=fluxcd
        .....
        - --feature-gates=DetectDrift=true
        - --feature-gates=CorrectDrift=false
```

and the helm-controller deployment's ServiceAccount just have the impersonate verb RBAC access granted. And per namespace have a  SA named `fluxcd` granted relevant per namespace scoped RBAC access. 

With above setup, a `HelmRelease` resource deployed without a `spec.serviceAccountName` field,  successfully gets installed in the first reconcile loop, but subsequent reconcile loops where it tries to detect drift, the diff client fails with below error (only relevant parts of the log included for brevity):
`
failed to diff release against cluster resources: [NetworkPolicy/ns1/allow-dashboard-ingress dry-run failed, reason: Forbidden: networkpolicies.networking.k8s.io "allow-dashboard-ingress" is forbidden: User "system:serviceaccount:flux-system:helm-controller" cannot patch resource "networkpolicies" in API group "networking.k8s.io" in the namespace "ns1"
`

Note that the diff client is falling back to helm-controller's SA instead of using the per namespace impersonator SA `fluxcd` set in the command-line flag.

This seems to be due to the differ client Impersonator setup isn't using the `github.com/fluxcd/helm-controller/internal/kube.DefaultServiceAccountName`  :

https://github.com/fluxcd/helm-controller/blob/v0.36.0/internal/controller/helmrelease_controller.go#L328

whereas all other Impersonator client usage in the helm-controller are using it correctly:

https://github.com/fluxcd/helm-controller/blob/v0.36.0/internal/controller/helmrelease_controller.go#L675
https://github.com/fluxcd/helm-controller/blob/v0.36.0/internal/kube/impersonate.go#L37

Any reasons for Differ Impersonator client setup not using the `--default-service-account` flag setting ?

ravilr:
cc @hiddeco  git blame shows the differ client was introduced in this commit https://github.com/fluxcd/helm-controller/commit/1240f201834afb02ce3e7b13dc2af74d8888f899

PR #774, opened 2023-09-18
controller: use `DefaultServiceAccount` in differ

hiddeco:
This addresses an issue in which the defunct `DefaultServiceAccount` from the `HelmReleaseReconciler` was being used to construct the impersonator used by the differ.

Fixes: #772


stefanprodan (approved):
LGTM

PR #818, opened 2023-11-24
Allow configuration of snapshot digest algorithm

hiddeco:
This introduces a `--snapshot-digest-algo` flag to allow configuring a different algorithm than SHA256.

This allows the user to for example configure `blake3`, which is potentially faster (and less resource intensive) on modern hardware.


stefanprodan (approved):
LGTM

PR #822, opened 2023-11-28
Correct cluster drift using patches

### LINKED ISSUE #805: Helm drift detection causes endless upgrade loop with CRDs
CR objects don't support StrategicMergePatch out of the box in Kubernetes, because their spec isn't builtin.
As a consequence of that, helm applies a MergePatch instead of StrategicMergePatch for CR objects.

The problem:
- Helm-controller drift detection triggers a helm upgrade when there are changes in CR objects
- If spec.upgrade.force is set to false (the default) on the helmrelease, this causes an endless loop because the helm upgrade will not effect any changes (effectively it just compares the previously applied configuration with the would-be-applied configuration, not taking into account the current cluster state

The (easier) solution:
- Helm-controller drift detection should take into account helm's behaviour when upgrading releases and not trigger an upgrade when spec.upgrade.force = false and the modified object is a CR only

Relevant change of helm's code (which was introduced because previously trying to patch CR objects was simply failing): https://github.com/helm/helm/pull/7269/files

hiddeco:
This changes the cluster drift correction behavior from performing a
Helm upgrade to performing create and patch API requests based on the
JSON Patch data.

Doing this is much lighter than performing a full release cycle, and
deals with the issue of Helm being unable to restore state of Custom
Resources without the `--force` flag being set. Which has unwanted
side-effects like forcing objects through a deletion/creation cycle.

After a drift correction attempt a Kubernetes Event is emitted, which
contains a summary of the created and patched resources, and a
collection of any (potential) errors.

As the goal is to restore state as best as we can, the drift correction
will be re-attempted until all resources have been restored to the
desired state.

Fixes #805

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,13 +15,14 @@ replace (
 )
 
 require (
+	github.com/fluxcd/cli-utils v0.36.0-flux.1
 	github.com/fluxcd/helm-controller/api v0.36.2
 	github.com/fluxcd/pkg/apis/acl v0.1.0
 	github.com/fluxcd/pkg/apis/event v0.6.0
 	github.com/fluxcd/pkg/apis/kustomize v1.2.0
 	github.com/fluxcd/pkg/apis/meta v1.2.0
 	github.com/fluxcd/pkg/runtime v0.43.0
-	github.com/fluxcd/pkg/ssa v0.34.0
+	github.com/fluxcd/pkg/ssa v0.35.0
 	github.com/fluxcd/pkg/testserver v0.5.0
 	github.com/fluxcd/source-controller/api v1.1.2
 	github.com/go-logr/logr v1.3.0
@@ -77,7 +78,6 @@ require (
 	github.com/evanphx/json-patch/v5 v5.7.0 // indirect
 	github.com/exponent-io/jsonpath v0.0.0-20210407135951-1de76d718b3f // indirect
 	github.com/fatih/color v1.13.0 // indirect
-	github.com/fluxcd/cli-utils v0.36.0-flux.1 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
 	github.com/go-errors/errors v1.4.2 // indirect
 	github.com/go-gorp/gorp/v3 v3.1.0 // indirect
--- a/internal/action/diff.go
+++ b/internal/action/diff.go
@@ -18,20 +18,27 @@ package action
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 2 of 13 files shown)


internal/action/diff.go
```diff
@@ -119,5 +124,125 @@ func Diff(ctx context.Context, config *helmaction.Configuration, rls *helmreleas
...
+		},
+		GroupVersion: obj.GetObjectKind().GroupVersionKind().Version,
+		Subject:      subjectFromObject(obj),
+		Action:       action,
+	}
+}
+
+// subjectFromObject returns a string representation of the given object.
+func subjectFromObject(obj client.Object) string {
+	return fmt.Sprintf("%s/%s/%s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetNamespace(), obj.GetName())
```
stefanprodan, line 207: I would use the SSA formatter that knows how to deal with non namespaced objects.

hiddeco: Can not use that here because I do not have a `object.ObjMetadata` at all times, but apply the same logic. Thanks :bow:

go.mod
```diff
@@ -15,13 +15,14 @@ replace (
...
 require (
+	github.com/fluxcd/cli-utils v0.36.0-flux.1
 	github.com/fluxcd/helm-controller/api v0.36.2
 	github.com/fluxcd/pkg/apis/acl v0.1.0
 	github.com/fluxcd/pkg/apis/event v0.6.0
 	github.com/fluxcd/pkg/apis/kustomize v1.2.0
 	github.com/fluxcd/pkg/apis/meta v1.2.0
 	github.com/fluxcd/pkg/runtime v0.43.0
-	github.com/fluxcd/pkg/ssa v0.34.0
+	github.com/fluxcd/pkg/ssa v0.34.1-0.20231127131230-ce912554b6d1
```
stefanprodan, line 25:
```suggestion
	github.com/fluxcd/pkg/ssa v0.35.0
```

stefanprodan (commented):
I've tested this with various custom resources changes and found no issues except for the `meta.helm.sh/release` annotations missing from objects which have been deleted and restored.

Fixed in: [0131f22](https://github.com/fluxcd/helm-controller/pull/822/commits/0131f2227b1e1f300046020378bd69b8ef412e3f)

stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅 

PS. I have tested this with various custom resources edits/deletions.

PR #844, opened 2023-12-13
loader: allow overwrite of URL hostname again

hiddeco:
This adds back the support for overwriting the host name a chart is downloaded from (again) using the `SOURCE_CONTROLLER_LOCALHOST` environment variable.


stefanprodan (approved):
LGTM

PR #845, opened 2023-12-13
loader: log HTTP errors to provide faster feedback

hiddeco:
Part of: https://github.com/fluxcd/flux2/issues/4472


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #848, opened 2023-12-13
Early stall condition detection after remediation

darkowlzz:
Detect stall condition due to exhausted remediation retry right after remediating. This helps return from the AtomicRelease.Reconcile() with proper stalled status condition and error. Without this, after remediation, especially during upgrade failures, a stalled condition detection required a new reconciliation, leaving the status of the object without any Reconciling or Stalled condition.


hiddeco (approved):
Tested this by hand and it works as intended, thank you @darkowlzz :1st_place_medal:

PR #850, opened 2023-12-14
controller: reset field managers for `v2beta1`

hiddeco:
Fix: https://github.com/fluxcd/flux2/issues/4479


stefanprodan (approved):
LGTM

Thanks @hiddeco I've tested this on Kubernetes 1.29.0 and after the upgrade a force reconcile is no longer needed.

The requirement is to bump the apiVersion and to to add an annotation to force the Kubernetes API to move the HelmRelease field managers to v2beta2. 

Example:
```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta2 # <- bump from v2beta1
kind: HelmRelease
metadata:
  annotations:
    fluxcd.io/upgrade: "v2beta2" # <- add annotation 
```

PR #861, opened 2023-12-21
Update Kubernetes dependencies

RafalKorepta:
Changes:

* Update k8s.io packages to v0.29.0

The restmapper.NewShortcutExpander changes required because of https://github.com/kubernetes/kubernetes/pull/117668. In `MemoryRESTClientGetter` the warning printer does not need printing warnings.


RafalKorepta:
Can I ask @stefanprodan or @hiddeco for review?

Thank you!

stefanprodan:
Updating the Kubernetes dependencies to the next minor is quite a complicated task which involves upgrading all the Flux packages in `fluxcd/pkg`, release them, then upgrade source-controller, release SC, then all the consumers like this controller. We also keep the Kubernetes client in sync with controller-runtime and cli-utils, once the upstream projects move to 1.29 we will start the upgrade procedure on all our repos.

Issue #855, opened 2023-12-18
Metallb installation with `driftDetection: mode: enabled` failed to apply revision


zaggash:
I'm trying to setup Metallb with this Kustomization:

```
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: apps-metallb
  namespace: flux-system
spec:
  path: /apps/metallb-system/metallb/app
  sourceRef:
    kind: GitRepository
    name: apps
    healthChecks:
      - apiVersion: helm.toolkit.fluxcd.io/v2beta2
        kind: HelmRelease
        name: metallb
        namespace: metallb-system
  interval: 30m
  retryInterval: 1m
  timeout: 3m
```

And this Helm Release:

```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta2
kind: HelmRelease
metadata:
  name: metallb
spec:
  interval: 15m
  driftDetection:
    mode: enabled
  chart:
    spec:
      chart: metallb
      version: "0.13.12"
      sourceRef:
        kind: HelmRepository
        name: metallb-charts
        namespace: flux-system
  maxHistory: 3
  install:
    createNamespace: true
    crds: CreateReplace
    remediation:
      retries: 3
  upgrade:
    cleanupOnFail: true
    crds: CreateReplace
    remediation:
      retries: 3
  uninstall:
    keepHistory: false
  values:
    controller:
      logLevel: warn
    speaker:
      logLevel: warn
    frr:
      enabled: false
```

`flux version`
```
flux: v2.2.0
distribution: flux-v2.2.1
helm-controller: v0.37.1
image-automation-controller: v0.37.0
image-reflector-controller: v0.31.1
kustomize-controller: v1.2.1
notification-controller: v1.2.3
source-controller: v1.2.3
```

`flux get kustomizations` is showing it is never Ready and marked as Unkown.

In the logs of the helm-controller I have

`k -n flux-system logs helm-controller-<id>`
```
[...]

PR #885, opened 2024-02-02
Update Ready condition during drift correction

darkowlzz:
Related to the issue described in https://github.com/fluxcd/flux2/issues/4524 and observations from https://github.com/fluxcd/helm-controller/issues/855. And another instance of the issue discussed on slack that revealed that stale Ready=False value results in such scenarios (described in #884 and more below).

This change builds on top of https://github.com/fluxcd/helm-controller/pull/884, to solve the issue at different levels. Please read #884 for more details about the issue.

Update the Ready condition during drift correction to reflect the current state of reconciliation. Without this, any previous Ready condition value continues to persist on the object. If there was a previous failure due to which Ready=False condition is present on the object, the same value continues to persist if the atomic release reconciliation enters a drift detection and correction loop. Resulting in the status to show inaccurate state of the reconciliation.

Examples of two different scenarios that arise due to this issue:
- If a release without any dependency is installed, the status shows Ready=True for InstallSucceeded reason. But right after the installation, if a drift is detected the status continues to show the same Ready=True value. There's no indication that a drift correction is going on in the status. The events and logs do show that drift correction is taking place. But it can be confusing to see positive Ready value. Also, since the Ready condition message is copied for Reconciling condition, Reconciling=True with a "Helm install succeeded..." is observed, refer https://github.com/fluxcd/helm-controller/blob/c2c1064a4cfda49115e561be33b7be78f06156ab/internal/reconcile/atomic_release.go#L222-L225. This scenario was reported in https://github.com/fluxcd/helm-controller/issues/855 and it results in the following status:
```yaml
status:
  conditions:
  - lastTransitionTime: "2023-12-19T11:22:43Z"
    message: Helm install succeeded for release metallb-system/metallb.v1 with chart
      metallb@0.13.12
    observedGeneration: 4
    reason: ProgressingWithRetry
    status: "True"
    type: Reconciling
  - lastTransitionTime: "2023-12-18T20:11:17Z"
    message: Helm install succeeded for release metallb-system/metallb.v1 with chart
      metallb@0.13.12
    observedGeneration: 1
    reason: InstallSucceeded
    status: "True"
    type: Ready
  - lastTransitionTime: "2023-12-18T20:11:17Z"
    message: Helm install succeeded for release metallb-system/metallb.v1 with chart
      metallb@0.13.12
    observedGeneration: 1
    reason: InstallSucceeded
    status: "True"
    type: Released
```

- If a release depends on another release, and reconciliation results in dependency not ready error at first, Ready=False condition is added on the object. On subsequent runs, even when the dependencies are ready, the Ready=False condition isn't updated, resulting in stale Ready value until atomic release reconciliation completes. But if the atomic reconciliation enters a drift detection and correction loop, the Ready=False with dependency error persists in the status. This gives the impression that something is wrong with dependency check but based on the logs and events, the controller could be stuck in drift detection and correction loop. This scenario was reported on slack and may also be the cause of the issue reported in https://github.com/fluxcd/flux2/issues/4524. This results in the following status:
```yaml
Status:
  Conditions:
    Last Transition Time:  2024-02-02T13:35:45Z
    Message:               dependency 'platform-flux-system/cilium' is not ready
    Observed Generation:   8
    Reason:                ProgressingWithRetry
    Status:                True
    Type:                  Reconciling
    Last Transition Time:  2024-02-01T15:31:45Z
    Message:               dependency 'platform-flux-system/cilium' is not ready
    Observed Generation:   6
    Reason:                DependencyNotReady
    Status:                False
    Type:                  Ready
    Last Transition Time:  2024-01-24T22:22:25Z
    Message:               Helm upgrade succeeded for release platform-external-secrets/external-secrets.v2 with chart external-secrets@0.9.11
    Observed Generation:   3
    Reason:                UpgradeSucceeded
    Status:                True
    Type:                  Released
```

Updating the Ready condition during drift detection shows the current state of reconciliation, avoiding the confusing scenarios described above. Following status is observed with the fix:

```yaml
status:
  conditions:
  - lastTransitionTime: "2024-02-02T23:01:02Z"
    message: correcting cluster drift
    observedGeneration: 1
    reason: ProgressingWithRetry
    status: "True"
    type: Reconciling
  - lastTransitionTime: "2024-02-02T23:00:37Z"
    message: correcting cluster drift
    observedGeneration: 1
    reason: Progressing
    status: Unknown
    type: Ready
  - lastTransitionTime: "2024-02-02T22:49:52Z"
    message: Helm install succeeded for release default/metallb.v1 with chart metallb@0.13.12
    observedGeneration: 1
    reason: InstallSucceeded
    status: "True"
    type: Released
```
The Ready=Unknown and Reconciling=True with their messages show correctly about the current state of reconciliation.

### DIFF
--- a/internal/reconcile/correct_cluster_drift.go
+++ b/internal/reconcile/correct_cluster_drift.go
@@ -24,6 +24,8 @@ import (
 	apierrutil "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
 
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/runtime/conditions"
 	"github.com/fluxcd/pkg/ssa"
 	"github.com/fluxcd/pkg/ssa/jsondiff"
 
@@ -63,6 +65,9 @@ func (r *CorrectClusterDrift) Reconcile(ctx context.Context, req *Request) error
 
 	ctx, cancel := context.WithTimeout(ctx, req.Object.GetTimeout().Duration)
 	defer cancel()
+
+	// Update condition to reflect the current status.
+	conditions.MarkUnknown(req.Object, meta.ReadyCondition, meta.ProgressingReason, "correcting cluster drift")
 
 	changeSet, err := action.ApplyDiff(ctx, r.configFactory.Build(nil), r.diff, r.fieldManager)
 	r.report(req.Object, changeSet, err)
--- a/internal/reconcile/correct_cluster_drift_test.go
+++ b/internal/reconcile/correct_cluster_drift_test.go
@@ -24,10 +24,13 @@ import (
 	. "github.com/onsi/gomega"
 	extjsondiff "github.com/wI2L/jsondiff"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apierrutil "k8s.io/apimachinery/pkg/util/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/runtime/conditions"
 	"github.com/fluxcd/pkg/ssa"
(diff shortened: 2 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

PR #907, opened 2024-03-06
Reintroduce missing events for helmChart reconciliation

souleb:
fixes [4649](https://github.com/fluxcd/flux2/issues/4649)

If implemented this PR reintroduces events for some failing actions during the reconciliation process, related to the `chart` retrieval and loading of `chart` and `values`.

We do not reintroduce all events (still available in logs), but where it is desirable to have notifications in place.


souleb:
Testing with a faulty configmap gives us:

```shell
{"level":"info","ts":"2024-03-06T11:09:39.678Z","logger":"event-server","msg":"dispatching event","eventInvolvedObject":{"kind":"HelmRelease","namespace":"backend","name":"redis","uid":"50f43354-d991-4149-bd7f-31074eed9a83","apiVersion":"helm.toolkit.fluxcd.io/v2beta2","resourceVersion":"625589"},"message":"could not resolve ConfigMap chart values reference 'backend/application-values-g9g5b268dh' with key 'values.yaml': key not found"}
```

```shell
Status:
  Conditions:
    Last Transition Time:  2024-03-06T11:10:17Z
    Message:               Fulfilling prerequisites
    Observed Generation:   8
    Reason:                Progressing
    Status:                True
    Type:                  Reconciling
    Last Transition Time:  2024-03-06T11:10:17Z
    Message:               could not resolve ConfigMap chart values reference 'backend/application-values-g9g5b268dh' with key 'values.yaml': key not found
    Observed Generation:   8
    Reason:                ValuesError
    Status:                False
    Type:                  Ready
```

internal/controller/helmrelease_controller.go
```diff
@@ -268,6 +269,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, patchHelpe
 
 		msg := fmt.Sprintf("could not get HelmChart object: %s", err.Error())
 		conditions.MarkFalse(obj, meta.ReadyCondition, v2.ArtifactFailedReason, msg)
+		r.Eventf(obj, eventv1.EventSeverityError, v2.ArtifactFailedReason, msg)
```
stefanprodan, line 272: I think this will spam is SC restarts and at upgrades. As all artifacts are removed from status, then re-added.

stefanprodan: I was referring to Flux upgrades

stefanprodan (approved):
LGTM

Thanks @souleb 🥇

internal/controller/helmrelease_controller.go
```diff
@@ -259,6 +259,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, patchHelpe
 			conditions.MarkStalled(obj, aclv1.AccessDeniedReason, err.Error())
 			conditions.MarkFalse(obj, meta.ReadyCondition, aclv1.AccessDeniedReason, err.Error())
 			conditions.Delete(obj, meta.ReconcilingCondition)
+			r.Eventf(obj, eventv1.EventSeverityError, aclv1.AccessDeniedReason, err.Error())
```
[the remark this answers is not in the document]
souleb: the way this was fixed in helm-controller was with this function

```go
func (r *HelmReleaseReconciler) event(_ context.Context, hr v2.HelmRelease, revision, severity, msg string) {
...
  eventType := corev1.EventTypeNormal
  if severity == eventv1.EventSeverityError {
    eventType = corev1.EventTypeWarning
  }
  r.EventRecorder.AnnotatedEventf(&hr, eventMeta, eventType, severity, msg)
}
```

and then in `fluxcd/pkg/runtime/events/recorder.go` we have this function:
```go
func eventTypeToSeverity(eventType string) string {
	switch eventType {
	case corev1.EventTypeWarning:
		return eventv1.EventSeverityError
	case eventv1.EventTypeTrace:
		return eventv1.EventSeverityTrace
	default:
		return eventv1.EventSeverityInfo
	}
}
```

This code makes sure that you always pass `warning` or `normal` to the recorder that can then re-switch to the right severity for NC.

For `Trace` events you can pass them as-is to the recorder that expects it, and do not send them to NC:
```go
	// Do not send trace events to notification controller,
	// traces are persisted as Kubernetes events only as normal events.
	if severity == eventv1.EventSeverityTrace {
		r.EventRecorder.AnnotatedEventf(object, annotations, corev1.EventTypeNormal, reason, messageFmt, args...)
		return
	}
```

I think we can make sure that the recorder infer the right `severity` and `final type` based on the passed `event type` directly in `fluxcd/pkg/runtime/events/recorder.go`. I'll give it a try.

PR #930, opened 2024-04-02
Add `notation` verification provider to API

stefanprodan:
Followup: https://github.com/fluxcd/source-controller/pull/1075


souleb:
we should add a test for it in `helmchart_template_test.go`.

stefanprodan:
> we should add a test for it in helmchart_template_test.go

We do have a test for the verify fields, duplicating the current test and changing the string value would not do anything as the kubebuilder validation doesn't run, `helmchart_template_test.go` uses a mock client.

souleb (approved):
LGTM! Thanks @stefanprodan

PR #935, opened 2024-04-09
Fix patching on drift detection

### LINKED ISSUE #925: [BUG] Helm drift detection on configmap patching '*** (after)' instead of the actual template from the helm chart
Hi Folks, 
we're running helmcontroller version v0.37.2 and enabled drift detection. Noticing that if any manual change was added to one of the configmaps of our helm chart, the helm controller detects a drifts and patches with a string and not the actual config map from the helm template.
```
{"level":"debug","ts":"2024-03-26T00:16:19.027Z","msg":"resource modified","controller":"helmrelease","controllerGroup":"helm.toolkit.fluxcd.io","controllerKind":"HelmRelease","HelmRelease":{"name":"<resource_name>","namespace":"flux-system"},"namespace":"flux-system","name":"<servicename>","reconcileID":"94f4e548-7273-44b3-a316-859516ec5dbe","resource":"ConfigMap/flux-system/<service>-config","patch":[{"value":"*** (after)","op":"replace","path":"/data/config.yaml"}]}
```
Config map after reconcile:
```
apiVersion: v1
data:
  config.yaml: '*** (after)'
kind: ConfigMap
metadata:
********
```
configmap template sample:
```
apiVersion: v1
data:
  config.yaml: |
    slaPeriod: 12h   # if this is manually updated to 6h, then drift detection runs
metadata:
********
```
the helmrelease also seems to be stuck with `ProgressingWithRetry`
```status:
  conditions:
  - lastTransitionTime: "2024-03-28T00:37:54Z"
    message: Helm upgrade succeeded for release flux-system/<chart_name>.v53 with chart <chart_name>@0.0.5
    observedGeneration: 21
    reason: ProgressingWithRetry
    status: "True"
    type: Reconciling
```

souleb:
fix #925 

Make sure that `/data` for configmaps are not masked

### DIFF
--- a/internal/reconcile/atomic_release.go
+++ b/internal/reconcile/atomic_release.go
@@ -382,8 +382,7 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state 
 					"resource", diff.ResourceName(change.DesiredObject))
 			case jsondiff.DiffTypeUpdate:
 				log.V(logger.DebugLevel).Info("resource modified",
-					"resource", diff.ResourceName(change.DesiredObject),
-					"patch", jsondiff.MaskSecretPatchData(change.Patch))
+					"resource", diff.ResourceName(change.DesiredObject))
 			}
 		}
 


stefanprodan (approved):
LGTM

Thanks @souleb 🏅

Issue #938, opened 2024-04-12
HelmRelease: CRDs of disabled subcharts get installed anyway


MBrouns:
### Describe the bug

Assume a helm chart with the following structure

```
.
├── Chart.yaml
├── charts
│   └── subchart
│       ├── Chart.yaml
│       ├── crds
│       │   └── crd.yaml
│       ├── templates
│       │   └── subchartpod.yaml
│       └── values.yaml
├── templates
│   └── mainchartpod.yaml
└── values.yaml
```

where the mainchart's Chart.yaml says:

```
dependencies:
  - name: subchart
    condition: subchart.enabled
```

When I create a helm release  with the following details (specifically note that the subchart is not enabled) 

```
apiVersion: helm.toolkit.fluxcd.io/v2beta2
kind: HelmRelease
metadata:
  name: fluxcd-test
  namespace: default
spec:
  interval: 10m
  timeout: 5m
  chart:
    spec:
      chart: mainchart
      version: "0.1.0"
      sourceRef:
        kind: GitRepository
        name: fluxcd-subchart-crd-test
      interval: 5m
  releaseName: crd-test
  values:
    replicaCount: 1
    subchart:
      enabled: false
```

I still see the crd from the subchart created in the cluster. This is only the case for the crd in the subchart, not the resources defined in the subchartpod.yaml. This is not what I expect, and also not what happens when installing the helm chart directly into the cluster

I have tested this for all combinations of Kubernetes 1.26.8, 1.27.8 and 1.28.3 and flux 2.2.0 and 2.2.3

Running `helm install test . --set 'subchart.enabled=false'` directly does not install the subcharts crd

### Steps to reproduce

1. install flux using the CLI
2. apply the following manifest to your cluster

```
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  name: fluxcd-subchart-crd-test
  namespace: default
spec:
  interval: 5m0s
  url: https://github.com/mbrouns/fluxcd-subchart-crd-test
  ref:
    branch: main
---
apiVersion: helm.toolkit.fluxcd.io/v2beta2
kind: HelmRelease
metadata:
  name: fluxcd-test
  namespace: default
spec:
  interval: 10m
  timeout: 5m
  chart:
    spec:
      chart: mainchart
      version: "0.1.0"
      sourceRef:
        kind: GitRepository
        name: fluxcd-subchart-crd-test
      interval: 5m
  releaseName: crd-test
  values:
[...]

PR #941, opened 2024-04-16
WIP: don't install CRDs of disabled subcharts

MBrouns:
Possible solution for #938

### DIFF
--- a/internal/action/install.go
+++ b/internal/action/install.go
@@ -55,6 +55,11 @@ func Install(ctx context.Context, config *helmaction.Configuration, obj *v2.Helm
 	if err != nil {
 		return nil, err
 	}
+
+	if err := helmchartutil.ProcessDependenciesWithMerge(chrt, vals); err != nil {
+		return nil, err
+	}
+
 	if err := applyCRDs(config, policy, chrt, setOriginVisitor(v2.GroupVersion.Group, obj.Namespace, obj.Name)); err != nil {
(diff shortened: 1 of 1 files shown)


internal/action/install.go
```diff
@@ -55,6 +55,11 @@ func Install(ctx context.Context, config *helmaction.Configuration, obj *v2.Helm
 	if err != nil {
 		return nil, err
 	}
+
+	if err := helmchartutil.ProcessDependenciesWithMerge(chrt, vals); err != nil {
+		return nil, err
+	}
```
stefanprodan, line 61: If this actually works, we'll need it on upgrade too.

souleb:
This results in `helmchartutil.ProcessDependenciesWithMerge` being called twice. We need a test case for idempotence here.

stefanprodan:
> We need a test case for idempotence here.

Reading this I hope we are safe:
- https://github.com/helm/helm/blob/main/pkg/chartutil/dependencies.go#L208-L211
- https://github.com/helm/helm/blob/main/pkg/chartutil/dependencies.go#L303-L315

In any case, if anywhere in that logic a map is used, calling the function twice will mess up the whole thing, so we really need a test.

internal/action/install.go
MBrouns: That makes sense. I should probably add it to `applyCRDs` instead to cover all the use-cases

stefanprodan:
Superseded by #1123

PR #968, opened 2024-05-07
Add the chart app version to status and events metadata

stefanprodan:
This PR adds the chart app version to the `.status.history` snapshot and sets the `app-version` annotation to all the issued events. 

This PR makes possible to expose `chart_app_version` as a label on the `gotk_resource_info` metric, ref: https://github.com/fluxcd/flux2-monitoring-example/pull/36

### DIFF
--- a/internal/reconcile/release.go
+++ b/internal/reconcile/release.go
@@ -200,11 +200,13 @@ func eventMessageWithLog(msg string, log *action.LogBuffer) string {
 // addMeta is a function that adds metadata to an event map.
 type addMeta func(map[string]string)
 
-// metaOCIDigestKey is the key for the OCI digest metadata.
-const metaOCIDigestKey = "oci-digest"
+const (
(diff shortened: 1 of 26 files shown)


relu (approved):
👍

internal/reconcile/release.go
```diff
@@ -203,6 +203,9 @@ type addMeta func(map[string]string)
 // metaOCIDigestKey is the key for the OCI digest metadata.
 const metaOCIDigestKey = "oci-digest"
 
+// metaAppVersionKey is the key for the app version found in chart metadata.
+const metaAppVersionKey = "app-version"
+
```
souleb, line 208:
```suggestion
const (
  metaOCIDigestKey = "oci-digest"
  // metaAppVersionKey is the key for the app version found in chart metadata.
  metaAppVersionKey = "app-version"
)

```

PR #965, opened 2024-05-05
Track changes in `.spec.postRenderers`

souleb:
fix #964

PR #972, opened 2024-05-08
PostRenderersDigest observation improvements

darkowlzz:
This adds a minor improvement to the implementation of the observing the post renderers digest in #965 by moving the observation set/update code from `summarize()` to atomic release reconciliation. `summarize()` is for summarizing the status conditions only and it is called by all the action sub-reconcilers. In addition, the post renderers digest value check is performed only for unprocessed new configuration. This is needed to prevent the atomic reconciliation from getting stuck in a loop when there are multiple upgrades, for example due to a config difference and a digest mismatch, and since the digest gets updated at the very end of a successful reconciliation, the release can get stuck. This make sure that the digest is checked only for unprocessed generation of the object. This behavior is only associated with the changes introduced here. Previously, the observation got updated early within the `summarize()` which got around such issue.

Also, adds more tests for various scenarios involving post renderers.

### DIFF
--- a/internal/reconcile/atomic_release.go
+++ b/internal/reconcile/atomic_release.go
@@ -38,7 +38,9 @@ import (
 	v2 "github.com/fluxcd/helm-controller/api/v2"
 	"github.com/fluxcd/helm-controller/internal/action"
 	"github.com/fluxcd/helm-controller/internal/diff"
+	"github.com/fluxcd/helm-controller/internal/digest"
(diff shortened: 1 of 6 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

souleb (approved):
LGTM! Thanks @darkowlzz

PR #905, opened 2024-03-05
Add support for `OCIRepository` as chartRef

souleb:
fixes #789 #903

PR #975, opened 2024-05-09
Prevent snapshot drift when upgrading to API v2

stefanprodan:
Mark `Observation.OCIDigest` as `omitempty` to avoid a snapshot drift when upgrading to API v2.

Before this fix, when updating the controller & CRDs from v0.x to v1.0 all releases were upgraded due to:
```
release not managed by controller: release not observed to be made for object
```

Whit this fix, the history snapshot remains unchanged, as the OCI digest empty string is no longer included in the calculation of the config digest.

Followup: https://github.com/fluxcd/helm-controller/pull/905

### DIFF
--- a/internal/release/digest_test.go
+++ b/internal/release/digest_test.go
@@ -37,7 +37,7 @@ func TestDigest(t *testing.T) {
 			rel: Observation{
 				Name: "foo",
 			},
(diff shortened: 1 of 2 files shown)


souleb (approved):
LGTM!

PR #980, opened 2024-05-10
Make chartTemplate a pointer in .spec.chart

souleb:
If implemented, this will omit empty `.spec.chart` in rendered yamls.


stefanprodan (approved):
LGTM

Thanks @souleb

seaneagan:
@souleb @stefanprodan was the breaking change to the v2beta2 types here intended? didn't see any mention of that being the case here or in the changelog

stefanprodan:
@seaneagan if you need v2beta2 stay on api v0.x. In v1.0.0 we have deprecated v2beta2, no one should use it with this version.

PR #982, opened 2024-05-10
Fix: Allow upgrading from v2beta1 to v2 (GA)

stefanprodan:
Add `.spec.chartRef` placeholder to v2beta1 to allow upgrading to v2 GA.


souleb (approved):
LGTM!

PR #990, opened 2024-05-14
Ignore 'v' version prefix in OCI artifact and Helm chart

makkes:
Tools such as Bitnami's charts-syncer strip the `v` prefix from the chart version so that the OCI artifact version differs from the version defined in the chart's metadata. This leads to an error similar to this returned from h-c:

```
artifact revision 1.14.5 does not match chart version v1.14.5
```

This PR makes h-c ignore a leading `v` prefix in either the chart version of the OCI artifact tag.

### DIFF
--- a/internal/controller/helmrelease_controller.go
+++ b/internal/controller/helmrelease_controller.go
@@ -915,7 +915,11 @@ func mutateChartWithSourceRevision(chart *chart.Chart, source sourcev1.Source) (
 	switch {
 	case strings.Contains(revision, "@"):
 		tagD := strings.Split(revision, "@")
-		if len(tagD) != 2 || tagD[0] != chart.Metadata.Version {
+		tagVer, err := semver.NewVersion(tagD[0])
+		if err != nil {
+			return "", fmt.Errorf("failed parsing artifact revision %s", tagD[0])
+		}
+		if len(tagD) != 2 || !tagVer.Equal(ver) {
 			return "", fmt.Errorf("artifact revision %s does not match chart version %s", tagD[0], chart.Metadata.Version)
 		}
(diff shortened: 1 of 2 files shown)


stefanprodan:
The `v` prefix is no valid for Helm charts, are you saying that you have a chart stored as `vX.Y.Z` in a Helm HTTP repo and charts-syncer removes it from Chart.yaml but not from the artifact tag?

makkes:
> The `v` prefix is no valid for Helm charts

I ran into this with cert-manager:

```
$ helm search repo jetstack/cert-manager                                                                                                                                                                                                                            
NAME                                    CHART VERSION   APP VERSION     DESCRIPTION                                                                                                                                                                                               
jetstack/cert-manager                   v1.14.5         v1.14.5         A Helm chart for cert-manager
```

They're using a v-prefixed version in their charts (that's also in the Chart.yaml).

> are you saying that you have a chart stored as `vX.Y.Z` in a Helm HTTP repo and charts-syncer removes it from Chart.yaml but not from the artifact tag?

The other way around: charts-syncer stores the Helm artifact in OCI tagged as `X.Y.Z` but the `Chart.yaml` defines it as `vX.Y.Z`.

stefanprodan:
Ok, we can backport this to v1.0.x but I would wait a week or two before doing a patch release.

stefanprodan:
We should also allow the reverse situation, v prefix in the tag, but no v in Chart.yaml.

makkes:
> We should also allow the reverse situation, v prefix in the tag, but no v in Chart.yaml.

It catches both. I implemented a test for each case.

stefanprodan (approved):
LGTM

Thanks @makkes 🏅

PR #1000, opened 2024-05-31
Move 'Deprecated' comments to own paragraph

hoffimar:
As described at https://go.dev/wiki/Deprecated, the `Deprecated` comments shall start a new paragraph. This way, the deprecation is also visible to other tools like staticcheck or vscode. Otherwise, one might get surprised by the removal of deprecated fields.
I also ran `make manifests`, which adds some newlines.


matheuscscp (approved):
LGTM! :rocket: 

Thanks @hoffimar, and sorry for taking so long to look into this, it's clearly a useful contribution :+1: 

I took the liberty to rebase and tidy things up instead of nudging you after so long :sweat_smile:

PR #1025, opened 2024-07-12
Fix incorrect use of format strings with the `conditions` package.

octo:
The `Mark…` functions in the `conditions` package accept a format string and (optional) arguments, just like `fmt.Printf` and friends.

In many places, the code passed an error message as the format string, causing it to be interpreted as a format string by the `fmt` package. This leads to issues when the message contains percent signs, e.g. URL-encoded values.

This PR adds a format string and shortens `err.Error()` to `err`, which yields the same output.

This change is identical in principle to fluxcd/source-controller#1529.

### DIFF
--- a/internal/controller/helmrelease_controller.go
+++ b/internal/controller/helmrelease_controller.go
@@ -252,7 +252,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, patchHelpe
 		if err := r.checkDependencies(ctx, obj); err != nil {
 			msg := fmt.Sprintf("dependencies do not meet ready condition (%s): retrying in %s",
 				err.Error(), r.requeueDependency.String())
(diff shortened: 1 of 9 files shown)


stefanprodan (approved):
LGTM

Thanks @octo

octo:
Thanks for the quick reviews @stefanprodan! 🎉

stefanprodan:
@octo please fix:
```
internal/reconcile/atomic_release.go:201:6: github.com/fluxcd/pkg/runtime/conditions.MarkStalled call needs 1 arg but has 2 args
```

octo:
> @octo please fix:
> 
> ```
> internal/reconcile/atomic_release.go:201:6: github.com/fluxcd/pkg/runtime/conditions.MarkStalled call needs 1 arg but has 2 args
> ```

✅ done

PR #1040, opened 2024-08-06
Upgrade dependenceis

r-darwish:
This fixes CVE-2024-6104 and CVE-2024-41110


stefanprodan:
We'll update all deps once Kubernetes 1.31 is released

PR #1048, opened 2024-08-15
Update dependencies to Kubernetes v1.31.0

stefanprodan:
Update dependencies to Kubernetes v1.31.0 and adapt the reconciler code to breaking changes in controller-runtime v0.19 and helm v3.15.4.

Part of: https://github.com/fluxcd/flux2/issues/4928


souleb (approved):
LGTM!

PR #1060, opened 2024-09-08
Allow overwriting inline values with targetPath

Preisschild:
Fix: https://github.com/fluxcd/flux2/issues/2330

When using Helm with both a values file and --set, the set will always have precedence over the values.yaml. This allows specifying the normal config inside a yaml file and specifying a secret outside that yaml via --set.

When using flux, this normally works too (using inline values to specify most of the config and a reference with targetPath to get a secret), but if you want to overwrite a secret key inside an array the inline values will always overwrite the whole array and the referenced secret key will not appear at all.

By making sure that when you specify a targetPath it has precedence over inline values it will be merged properly.

The Workaround mentioned in the original issue is limiting, because you can not easily use kustomize to overwrite certain keys in a configmap values.yaml stringBlock (for example in a multicluster setup with a common base), but you can do it with inline values.


hiddeco (approved):

stefanprodan (approved):
LGTM

Thanks @Preisschild 🏅

PR #1084, opened 2024-10-08
fix: handle "leader changed" errors

luisdavim:
Fixes https://github.com/fluxcd/flux2/issues/4804  by copying the solution used in helm/helm#11426


luisdavim:
see also https://github.com/helm/helm/issues/13052

hiddeco (approved):

luisdavim:
I see there's a new helm tag that includes the changes to make this functionality public, should I update this PR to bump the helm dependency and use that to avoid the extra code copied from there?

stefanprodan:
> should I update this PR to bump the helm dependency and use that to avoid the extra code copied from there?

Yes please

luisdavim:
This is strange, the file is here: https://github.com/helm/helm/blob/main/pkg/kube/roundtripper.go and the latest tag is from 2 days ago, well after that file was added but I don't see the round tripper in https://pkg.go.dev/helm.sh/helm/v3@v3.16.3/pkg/kube and even after updating the impoort, go says the type doesn't exist... I even cleaned my mod cache to be sure.

luisdavim:
Ok, I see the issue, my change to make the round tripper public was not picked into the `release-3.16` branch and the tag is from there, that's why the type is not available :( for now we'd have to keep the copy of the round tripper code, would it be ok to include this on the next release? we run k8s on the edge with limited resources and run into this issue a few times...

hiddeco:
This sounds good to me. If you can please add a reference to the PR with a `// TODO` that we should change this next minor Helm release, we won't forget about it.

Thank you for your patience and effort.

luisdavim:
> This sounds good to me. If you can please add a reference to the PR with a `// TODO` that we should change this next minor Helm release, we won't forget about it.
> 
> Thank you for your patience and effort.

I to have a TODO comment there already but links to the issue, I can also add a link to the PR.

Issue #1099, opened 2024-10-31
OCI artifact semver alignment with Helm for tags with underscore


baburciu:
[`HelmRelease` referencing a chart directly from an `OCIRepository` ](https://fluxcd.io/flux/guides/helmreleases/#using-a-chart-reference)  fails if the `OCIRepository.spec.ref.tag` contains the underscore (`_`) character.

Example:
```yaml
  apiVersion: helm.toolkit.fluxcd.io/v2
  kind: HelmRelease
  metadata:
    :
    name: monitoring-crd
    namespace: sylva-system    
    :
  spec:
    chartRef:
      kind: OCIRepository
      name: unit-monitoring-crd
    install:
      createNamespace: true
    interval: 15m
    :
  status:
    conditions:
    :
    - lastTransitionTime: "2024-10-25T22:07:45Z"
      message: artifact revision 104.1.1_up57.0.3 does not match chart version 104.1.1+up57.0.3
      observedGeneration: 2
      reason: ChartMutateError
      status: "False"
      type: Ready
---
  apiVersion: source.toolkit.fluxcd.io/v1beta2
  kind: OCIRepository
  metadata:
    :
    name: unit-monitoring-crd
    namespace: sylva-system
    :
  spec:
    interval: 60m0s
    provider: generic
    ref:
      tag: 104.1.1_up57.0.3
    timeout: 60s
    url: oci://registry.gitlab.com/sylva-projects/sylva-core/rancher-monitoring-crd
  status:
    artifact:
      :
      revision: 104.1.1_up57.0.3@sha256:85ea0013056258452ce27431392167ac036fd7fd1cb64df3a750b04851ac6d20
``` 

This happens due to https://github.com/fluxcd/helm-controller/blob/v1.0.1/internal/controller/helmrelease_controller.go#L917-L920 check, which expects the _tag_ from the _tag@digest_ formatted OCI artifact manifest digest (which [ends up as the `OCIRepository.status.artifact.revision`](https://github.com/fluxcd/source-controller/blob/v1.3.0/api/v1beta2/ocirepository_types.go#L256-L260)), to be equal to the `Chart.yaml`-defined version.

[...]

PR #1102, opened 2024-10-31
Replace `_` with `+` when verifying the chart version matches the OCI artifact tag

baburciu:
Closes https://github.com/fluxcd/helm-controller/issues/1099

OCI tags are not semver compatible, with `+` character not allowed. Helm worked around this by converting `+` to `_` before pushing to the OCI repo and vice versa when pulling from the repo.

This PR makes helm-controller adhere to that convention, by converting OCI artifact tag using `_` to `+` replacement prior to any check or processing.

### DIFF
--- a/internal/controller/helmrelease_controller.go
+++ b/internal/controller/helmrelease_controller.go
@@ -916,9 +916,12 @@ func mutateChartWithSourceRevision(chart *chart.Chart, source sourcev1.Source) (
 	switch {
 	case strings.Contains(revision, "@"):
 		tagD := strings.Split(revision, "@")
-		tagVer, err := semver.NewVersion(tagD[0])
+		// replace '+' with '_' for OCI tag semver compatibility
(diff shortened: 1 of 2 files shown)


baburciu:
I'm not sure what should be best returned as error
```go
return "", fmt.Errorf("failed parsing artifact revision %s", tagD[0])
```
or
```go
return "", fmt.Errorf("failed parsing artifact revision %s", tagConverted)
```
but the conversion needs to happen prior to calling `semver.NewVersion`

stefanprodan:
I think we should return the one with a `+` as that's the string we try to validate as semver. 

PS. This PR needs a test using a tag that has `_`.

baburciu:
thank you, I've added a test for it, which is passing `make test`. I'm new to this, so if it's not that good, please inform

stefanprodan (approved):
LGTM

Thanks @baburciu 🥇

PR #1111, opened 2024-11-18
fix: Add back the helm suspended metric

jmleddy:
At some point we had this and then we lost it. Discovered after we started suspending a bunch of things but could not get this metric to appear, meaning we are currently in a quasi-state of releases suspended across all our clusters that we don't know about.

### DIFF
--- a/internal/controller/helmrelease_controller.go
+++ b/internal/controller/helmrelease_controller.go
@@ -205,8 +205,9 @@ func (r *HelmReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			log.Error(err, "failed to wait for object to sync in-cache after patching")
 		}
 
-		// Record the duration of the reconciliation.
+		// Record Prometheus metrics.
 		r.Metrics.RecordDuration(ctx, obj, start)
+		r.Metrics.RecordSuspend(ctx, obj, obj.Spec.Suspend)
 	}()
 
 	// Examine if the object is under deletion.


stefanprodan:
The suspend label is has been included in the `gotk_resource_info` which is provided by kube-state-metrics. I recommend you migrate your alerts and dashboard to the new metric as the old ones have been deprecated long ago. Docs here: https://fluxcd.io/flux/monitoring/metrics/#resource-metrics

jmleddy:
This is extra toil for us to get a back metric that we were alerting on and have completely lost. We have no idea how many helm releases are paused not applying resource request updates or whatever. And it's inconsistently applied. Why do our kustomizations still report when they are stalled but our helm releases don't. I realize that there are different maintainers that  have different opinions about what metrics should be exposed, but to the end user this all just looks like "flux", since all the controllers come with flux.

For anyone that might find this PR and wonder what the kube-state-metrics config is, seems to be [here](https://github.com/fluxcd/flux2-monitoring-example/blob/main/monitoring/controllers/kube-prometheus-stack/kube-state-metrics-config.yaml)

jmleddy:
Also there isn't any place in the `kube-prometheus-stack` helm chart to add this information. The barrier for fixing this is relatively high https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack?modal=values&path=kube-state-metrics

stefanprodan:
>  I realize that there are different maintainers that have different opinions

The core maintainers make the decisions for the common behaviour of Flux controllers and metrics fall into this category. We made the decision to drop the resource specific metrics from the controllers exporters and rely on kube-prometheus-stack. The deprecation notice can be found here: https://fluxcd.io/flux/monitoring/metrics/#warning-deprecated-resource-metrics

This controller was last promoted to GA, so we removed the deprecated metrics from it, but we should've done that in all controllers. We'll make sure the old metrics are removed in the next release across all Flux components.

jmleddy:
Thank you, though I would prefer you add back this metric everywhere to avoid requiring everyone to add 275 lines of yaml to their kube-prometheus-stack helm chart, the inconsistency is even worse than the first decision as it feels uneven. And probably also led us to slower detection of the issue as it was still finding suspends "sometimes". So looking forward to having a consistent view from the Flux controller maintainers here :)

jmleddy:
Also, is this you? https://github.com/fluxcd/flux2-monitoring-example/issues/35#issuecomment-2226364651

jmleddy:
Okay I thought the controller was running as part of the operator I must not have my kube config right. I'll run it as part of the  ksm thanks!

jmleddy:
Okay I thought the controller was running as part of the operator I must not have my kube config right. I'll run it as part of the ksm thanks!

PR #1181, opened 2025-02-24
Revert controller-runtime to v0.19

matheuscscp:
Revert controller-runtime from v0.20 to v0.19.  The Kubernetes [Aggregated Discovery](https://github.com/kubernetes/enhancements/issues/3352)  was enabled in [controller-runtime#2901](https://github.com/kubernetes-sigs/controller-runtime/pull/2901) and looks like this feature is buggy resulting in old API versions being returned as the preferred version randomly. 

This was observed on lots of cluster by Flux users, including on Kubernetes 1.32, 1.31 and OpenShift 4.16.

xref: https://github.com/fluxcd/kustomize-controller/pull/1374

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,11 @@ replace (
 	sigs.k8s.io/kustomize/kyaml => sigs.k8s.io/kustomize/kyaml v0.19.0
 )
 
+// Pin controller-runtime to v0.19.6
+// The Aggregated Disovery is buggy resulting in old API versions being returned as the preferred version.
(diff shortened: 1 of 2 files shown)


stefanprodan:
I realised this is not needed, helm-controller does not uses the RestMapper from controller-runtime, it creates its own with client-go https://github.com/fluxcd/helm-controller/blob/25c6bb691df30a3c2fd8a4b19ce65bc134158c90/internal/kube/client.go#L170

PR #1188, opened 2025-03-03
fix: fix the wrong error return value

drawdrop:
fix the wrong error return value


matheuscscp (approved):

matheuscscp:
Thanks very much @drawdrop :pray:

PR #1212, opened 2025-05-07
Add `DisableChartDigestTracking` feature gate

stefanprodan:
Fix: https://github.com/fluxcd/flux2/issues/4910

When using `OCIRepository` for Helm charts that set the chart version as a label e.g. `app.kubernetes.io/version: {{ .Chart.Version }}`, helm-controller will fail to install the Helm release due to appending the digest to the chart version:

```
Helm upgrade failed  invalid: spec.template.labels: Invalid value: "0.9.3+4fda46fd8c4e": a valid label must be an empty string or consist of alphanumeric characters
```

This PR makes it possible to disable the chart digest tracking at controller level. When the feature gate is enabled, Flux will no longer trigger a Helm release upgrade if the digest changed unless the chart version changes too.

To deploy the release candidate and test the `DisableChartDigestTracking` feature gate, add the following patch to the flux-system overlay:

```yaml
patches:
  - target:
      kind: Deployment
      name: "helm-controller"
    patch: |
      - op: add
        path: /spec/template/spec/containers/0/args/-
        value: --feature-gates=DisableChartDigestTracking=true
      - op: replace
        path: /spec/template/spec/containers/0/image
        value: ghcr.io/fluxcd/helm-controller:rc-da305300
```

### DIFF
--- a/docs/spec/v2/helmreleases.md
+++ b/docs/spec/v2/helmreleases.md
@@ -217,6 +217,9 @@ The controller will automatically perform a Helm upgrade when the `OCIRepository
 detects a new digest in the OCI artifact stored in registry, even if the version
 inside `Chart.yaml` is unchanged.
 
+**Note:** Disabling the appending of the digest to the chart version can be done
+with the `--feature-gates=DisableChartDigestTracking=true` controller flag.
+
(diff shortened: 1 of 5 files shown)


matheuscscp (approved):
LGTM

Issue #1217, opened 2025-05-16
Add CommonLabels to HelmRelease kustomize postRender


abhijith-darshan:
Currently the `postRender` in `HelmRelease` only supports `patches` and `images` fields for `kustomize`.

~~Introduce `commonLabels` and `commonAnnotations` keys supported by `kustomize` to allow setting some common labels and keys across all resources.~~

This reduces the overhead of prior knowledge of knowing the `apiVersion`, `kind`, resource naming conventions coming in from the helm chart.

This would be a huge benefit for platform teams that provide kubernetes resource abstractions by using flux underneath. It allows platform specific conventions to be applied on public Helm charts.

Update:

The existing convention of `CommonMetadata` in `kustomize-controller` is re-used here in `.spec` of `HelmRelease`  instead of adding `commonLabels` and `commonAnnotations` to `.spec.postRenderers[*].Kustomize`

abhijith-darshan:
I have opened a PR #1223 for this issue as well.

Considering the following spec -

```yaml
spec:
  chart:
    spec:
      chart: podinfo
      interval: 10m
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: podinfo
      version: '>=6.0.0 <7.0.0'
  interval: 5m
  postRenderers:
    - kustomize:
        commonAnnotations:
          fluxcd.io/foo: bar
        commonLabels:
          fluxcd.io/bar: baz
    - kustomize:
        commonAnnotations:
          fluxcd.io/bar: boo
          fluxcd.io/foo: baz # duplicate key
```

This should propagate common annotation and labels to all resources in the helm chart.

In the scenario where duplicate label keys or annotation keys are used in multiple kustomize `postRenderers` the last one wins since it is executed in order.

stefanprodan:
Implemented in https://github.com/fluxcd/helm-controller/pull/1223

PR #1223, opened 2025-05-19
feat: adds CommonMetadata to HelmRelease postrenderer

abhijith-darshan:
This PR addresses Issue #1217 to pass common labels and common annotations to `postRender` kustomize

Following the existing convention of `CommonMetadata` that is used in `kustomize-controller` to add labels and annotations, `CommonMetadata` is added to `HelmRelease` spec so that a common set of annotations and labels can be propagated to all resources in the helm chart without having to know the values schema of the chart


abhijith-darshan:
@hiddeco any further recommendations for this PR?

api/v2/helmrelease_types.go
```diff
@@ -185,6 +185,14 @@ type ValuesReference = meta.ValuesReference
 
 // Kustomize Helm PostRenderer specification.
 type Kustomize struct {
+	// Common labels to set on all resources.
+	// When duplicate labels are found across postRenderers, the last one wins.
+	// +optional
+	CommonLabels map[string]string `json:"commonLabels,omitempty"`
```
hiddeco: I wonder if it might be better to expose the full options available in Kustomize, rather than giving the false impression this makes use of the deprecated `commonLabels` feature.

stefanprodan: We have `commonMetadata` in Flux Kustomization that doesn’t alter the label selectors. I would go with this option instead of the deprecated commonLabels.

abhijith-darshan: I saw this PR - https://github.com/fluxcd/kustomize-controller/pull/817

> The main difference to the Kustomize [commonLabels](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/commonlabels/) and [commonAnnotations](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/commonannotations/), is that the controller sets the labels and annotations only to the top level metadata field, without patching the Kubernetes Deployment spec.template or the Service spec.selector.

but in this case it would be useful to propagate the information to even the annotations or labels of the pods created by `Deployment` 

As this is not the use case of flux kustomize resource I would request to allow this option in the helm postRenderer.

How about I change this to `commonMetadata` format as suggested but expose further boolean option to `IncludeTemplates`?

Would that be acceptable?

stefanprodan: Changing the label selectors will always result in a Helm upgrade failure with an immutable field error. I don't see why would we expose a way for people to shoot themselves in the foot.

abhijith-darshan: @stefanprodan - sorry I don't mean Deployment's `.spec.selector.matchLabels` and you are right they are immutable...

I meant Deployment's `template.metadata.labels` which can be modified and that is what kustomize `IncludeTemplates` does.

and I did it this way -

```go
	if k.CommonLabels != nil {
		cfg.Labels = make([]kustypes.Label, 0)
		cfg.Labels = append(cfg.Labels, kustypes.Label{
			Pairs:            k.CommonLabels,
			IncludeSelectors: false,
			IncludeTemplates: true,
		})
	}
```

abhijith-darshan: With this I did not face any issues in `HelmRelease`, so I was asking if this can be exposed in `postRenderer` spec

Also what is recommended here? to import the `commonMetadata` types from `kustomize_controller` or be independent of it?

abhijith-darshan:
> @abhijith-darshan the `commonMetadata` should be added to `.spec`, same as we do for Flux Kustomization. Please copy the type instead of adding a dependency to kustomize-controller. As for the implementation, we already have code that adds labels to all resources to track the HelmRelease owner, we can extend that code to add the extra metadata.

As suggested - 

- `commonMetadata` from `.spec.postRenderers` to `.spec`.
- removed the dependency to `kustomize-controller` and copied over the types for `CommonMetadata`
- merged the logic of `origin_labels.go` into `common_renderer.go` 
- introduced `transformers.go` to re-use _kustomize builtins_ `LabelTransformer` and `AnnotationsTransformer`
- `BuildPostRenderers` gives precedence to `OriginLabels` rendering over `CommonMetadata` in case of label key collision (unlikely to happen but the edge case is covered)
- Updated tests but please let me know if I have missed something

internal/postrender/common_renderer_test.go
```diff
@@ -0,0 +1,190 @@
+/*
+Copyright 2021 The Flux authors
```
stefanprodan:
```suggestion
Copyright 2025 The Flux authors
```

internal/postrender/transformer.go
```diff
@@ -0,0 +1,55 @@
+package postrender
```
stefanprodan: Copyright header is missing

stefanprodan (commented):
Please add a section to the docs about common metadata before https://github.com/fluxcd/helm-controller/blob/main/docs/spec/v2/helmreleases.md#post-renderers

api/go.mod
```diff
@@ -21,9 +21,9 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/spf13/pflag v1.0.6 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
-	golang.org/x/net v0.39.0 // indirect
-	golang.org/x/text v0.24.0 // indirect
-	golang.org/x/tools v0.32.0 // indirect
+	golang.org/x/net v0.40.0 // indirect
```
stefanprodan: Please undo all changes to go.mod and go.sum in the api and the main ones

abhijith-darshan: done in [132abe9](https://github.com/fluxcd/helm-controller/pull/1223/commits/132abe998801a8b45128f364326c4be68e81a9a7)

internal/postrender/transformer.go
abhijith-darshan: Done in [8576fb5](https://github.com/fluxcd/helm-controller/pull/1223/commits/8576fb56608829bc648258dc9d90dfa0f2e73424)

stefanprodan:
@abhijith-darshan can you please squash the commits

PR #1271, opened 2025-07-22
Extend the readiness evaluation of dependencies with CEL expressions

stefanprodan:
This PR extends the HelmRelease API with the `.spec.dependsOn[].readyExpr` optional field that can be used to define a CEL expression to determine the readiness of a HelmRelease dependency. 

This is helpful for when custom logic is needed to determine if a dependency is ready. For example, when performing a lockstep upgrade, the `readyExpr` can be used to verify that a dependency has a matching version in values before proceeding with the reconciliation of the dependent HelmRelease.

```yaml
kind: HelmRelease
metadata:
  name: backend
  namespace: default
spec:
  # ...omitted for brevity
  values: 
    app:
      version: v1.2.3
---
kind: HelmRelease
metadata:
  name: frontend
  namespace: default
spec:
  # ...omitted for brevity
  values:
    app:
      version: v1.2.3
  dependsOn:
    - name: backend
      readyExpr: >
        dep.spec.values.app.version == self.spec.values.app.version &&
        dep.status.conditions.filter(e, e.type == 'Ready').all(e, e.status == 'True') &&
        dep.metadata.generation == dep.status.observedGeneration
```

The CEL expression contains the following variables:

- `dep`: The dependency HelmRelease object being evaluated.
- `self`: The HelmRelease object being reconciled.

**Note:** When `readyExpr` is specified, the built-in readiness check is replaced by the logic defined in the CEL expression. You can configure the controller to run both the CEL expression evaluation and the built-in readiness check, with the `AdditiveCELDependencyCheck` [feature gate](https://fluxcd.io/flux/components/helm/options/#feature-gates).

xref: https://github.com/fluxcd/kustomize-controller/pull/1491 & https://github.com/fluxcd/kustomize-controller/issues/1490


matheuscscp (approved):
LGTM! 🚀

Issue #1278, opened 2025-08-07
Introduce `RetryOnFailure` lifecycle management strategy


stefanprodan:
This is a proposal for introducing a new way of performing lifecycle management akin to Flux kustomize-controller operational mode.

The `RetryOnFailure` strategy is suitable for statefulsets and other workloads that cannot tolerate rollbacks and have a high rollout duration susceptible to health check timeouts and transient capacity errors.

The `RetryOnFailure` strategy will ensure that:

- An installation failure will not be retried immediately, instead, the controller will retry with an upgrade at a fixed interval defined in the HelmRelease.
- An upgrade failure will leave the Helm release in a failed state without performing any remediations.
- An upgrade failure will never result in a rollback or uninstall.
- Upgrade failures are always retried at a fixed interval defined in the HelmRelease.

## API Changes

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
spec:
  install:
    strategy:
      name: RetryOnFailure # defaults to RemediateOnFailure
      retryInterval: 5m # only used for RetryOnFailure (defaults to 5m)
    remediation: # ignored for RetryOnFailure
      retries: 2
  upgrade:
    strategy:
      name: RetryOnFailure # defaults to RemediateOnFailure
      retryInterval: 5m # only used for RetryOnFailure (defaults to 5m)
    remediation: # ignored for RetryOnFailure
      retries: 2
      strategy: rollback
```

When not specified, or when the strategy is set to `RemediateOnFailure`, the lifecycle management works like before.

For installations, the `RetryOnFailure` strategy will perform an uninstall on failure, then will rerun the installation after the specified retry interval.

For upgrades, the `RetryOnFailure` strategy will behave like `flux reconcile hr --force` when the remediation retries are set to 0.

matheuscscp:
> the controller will uninstall the release

Is an uninstall strictly required? I was able to trigger a successful *upgrade* after a `failed` install. But not a successful *install*, as the Helm SDK errors out with `cannot re-use a name that is still in use`. If switching to upgrades after a `failed` install is an acceptable behavior, we can avoid uninstalls.

matheuscscp:
[...]

PR #1281, opened 2025-08-08
Introduce `RetryOnFailure` lifecycle management strategy

matheuscscp:
Closes: #1278

TODO:
* [x] Automatic tests.

This PR implements the API enhancements described in #1278.

I tested this thoroughly, including in the scenario of pods in `ImagePullBackOff` due to unexisting image tags. When the tags are pushed, the HelmRelease object eventually converges without manual intervention.

The behavior of the CLI commands `flux reconcile hr` and `flux reconcile hr --force` is the same for a HelmRelease in the `failed` state: a release action is attempted. If the HelmRelease is Ready, i.e. in the `deployed` state, the behaviors of these two commands are not the same, but they match the behaviors for a HelmRelease object that is not using this feature. In this case, the behavior is no-op for `flux reconcile hr`, and perform a release action for `flux reconcile hr --force`.

Because the retry strategies cause a time-based requeue (and do not end in a terminal state), running the `flux reconcile hr` command when the release is `failed`, with or without `--force`, results in an immediate release action, and *does not remove* any in-flight `RequeueAfter` operations. In particular, if an in-flight `RequeueAfter` completes during a reconciliation, another reconciliation will run immediately after. I observed this many times due to using a very short `.retryInterval` for speeding up my tests.

Another common behavior that I noticed while testing the *install* strategy is that, when the source object (e.g. OCIRepository) is created together with the HelmRelease, two reconciliations are processed: the one produced by "requeue dependency" and the watch event from the source object becoming Ready. This results in the first retry happening immediately after the first attempt, i.e. in two consecutive failed installs.

It's also important to note that using the install strategy causes the transition of `.status.lastAttemptedReleaseAction` from `install` to `upgrade`. So if this upgrade fails, the *upgrade* strategy configuration will now be the one driving the behavior, the install strategy configuration will no longer be at play. Therefore, to successfully achieve a HelmRelease that is automatically retried without any remediations, the spec must look like this:

```yaml
kind: HelmRelease
spec:
  install:
    strategy: # the install strategy will cause the first retry, which will be an upgrade (the first one)
      name: RetryOnFailure
  upgrade:
    strategy: # the upgrade strategy will cause the subsequent retries, which will all be upgrades
      name: RetryOnFailure
```


internal/controller/helmrelease_controller.go
```diff
@@ -392,10 +392,12 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context, patchHelpe
 		Chart:  loadedChart,
 		Values: values,
 	}); err != nil {
-		if errors.Is(err, intreconcile.ErrMustRequeue) {
+		switch {
+		case errors.Is(err, intreconcile.ErrRetryAfterInterval):
+			return ctrl.Result{RequeueAfter: obj.GetActiveRetry().Duration}, nil
```
stefanprodan:
```suggestion
			return jitter.JitteredRequeueInterval(ctrl.Result{RequeueAfter: obj.GetActiveRetry().Duration}), nil
```

stefanprodan (commented):
I suggest we apply jitter for the retries to spread out the reconciliations

internal/controller/helmrelease_controller.go
matheuscscp: Fixed, thanks!

internal/reconcile/atomic_release.go
```diff
@@ -278,21 +290,23 @@ func (r *AtomicRelease) Reconcile(ctx context.Context, req *Request) error {
...
+				}
+
 				remediation := req.Object.GetActiveRemediation()
 				if remediation == nil || !remediation.RetriesExhausted(req.Object) {
 					conditions.MarkReconciling(req.Object, meta.ProgressingWithRetryReason, "%s", conditions.GetMessage(req.Object, meta.ReadyCondition))
 					return ErrMustRequeue
 				}
-				// Check if retries have exhausted after remediation for early
-				// stall condition detection.
-				if remediation != nil && remediation.RetriesExhausted(req.Object) {
```
matheuscscp, line 288: This check is a tautology, it's basically the negation of the check right above. I removed it in this separate commit: https://github.com/fluxcd/helm-controller/pull/1281/commits/079ae1b7e7a6e307a29b582afb095323749ac5c6

matheuscscp (commented):
After deeply understanding the state machine, it turns out that it's simpler to introduce all the three action strategies in the same PR.

internal/reconcile/atomic_release.go
```diff
@@ -266,6 +278,14 @@ func (r *AtomicRelease) Reconcile(ctx context.Context, req *Request) error {
 			// Run the action sub-reconciler.
 			log.Info(fmt.Sprintf("running '%s' action with timeout of %s", next.Name(), timeoutForAction(next, req.Object).String()))
 			if err = next.Reconcile(ctx, req); err != nil {
+				if retry := req.Object.GetActiveRetry(); retry != nil {
+					log.Error(err, fmt.Sprintf("failed to run '%s' action", next.Name()))
```
matheuscscp, line 282: Here, this error was already recorded in a condition and was sent in Kubernetes and notification-controller events. But no logs show up because we are handling it here and returning a different error, so I'm logging it here

matheuscscp:
I tested this PR in the following scenarios:

1. Helm tests fail.
2. Chart build fails due to templating error. In this case, the `HelmReleaseReconciler` was returning the error to `controller-runtime`, resulting in retries with exponential backoff. To fix this I applied this [diff](https://github.com/fluxcd/helm-controller/compare/7f2a6e212f367c8dfa3a386ff76ff8a4b4be82fc..b6d1e3c0ca71814249ce69d2cfe4566ec7060faf).
3. Build works, but YAML apply fails. I tested this by making the chart apply a resource in a namespace it didn't have RBAC for.
4. Install succeeds, then I delete the Helm storage secret.

In all the cases above, after fixing the issue in scenario 2, the controller behavior matches the expectations:
* Retries after the configured and appropriate `.retryInterval`.
* A status condition is updated with the error.
* A log is emitted with the error (after applying this [diff](https://github.com/fluxcd/helm-controller/compare/b6d1e3c0ca71814249ce69d2cfe4566ec7060faf..2bb4699dedd2006f16a9086f6bc1d9158e5f2aa8)).
* A Kubernetes and notification-controller event is sent with the error.

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1282, opened 2025-08-13
Record the last Helm release action duration in status

stefanprodan:
This PR makes the helm-controller report the duration of the last Helm release action it attempted to perform in the `.status.lastAttemptedReleaseActionDuration` field. Recording the install or upgrade duration information can be useful for debugging purposes.


matheuscscp (approved):
LGTM! 🚀

PR #1284, opened 2025-08-17
[RFC-0010] Add multi-tenancy lockdown for kubeconfig

cappyzawa:
Part of: https://github.com/fluxcd/flux2/issues/5465


internal/controller/helmrelease_controller.go
```diff
@@ -732,7 +732,10 @@ func (r *HelmReleaseReconciler) buildRESTClientGetter(ctx context.Context, obj *
 	var err error
 	switch {
 	case kc.ConfigMapRef != nil && kc.SecretRef == nil:
-		var opts []auth.Option
+		opts := []auth.Option{
+			auth.WithClient(r.Client),
```
matheuscscp: No need, `authutils.GetRESTConfig()` already does this :+1:

cappyzawa: That’s true…!

matheuscscp (approved):
LGTM! :rocket:

PR #1293, opened 2025-09-05
[RFC-0012] Add support for `ExternalArtifact` source type to `chartRef`

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/5504


internal/controller/helmrelease_controller.go
```diff
@@ -832,13 +867,20 @@ func (r *HelmReleaseReconciler) waitForHistoryCacheSync(obj *v2.HelmRelease) wai
 }
 
 func isSourceReady(obj sourcev1.Source) (bool, string) {
+	if o, ok := obj.(*sourcev1.ExternalArtifact); ok {
+		if obj.GetArtifact() == nil {
```
matheuscscp, line 871: Is this not also checked for the other SC APIs somewhere? It's kinda weird to be type-asserting only one of the APIs here

internal/test/external_artifact_test.go
matheuscscp, line 1: Any particular reason for introducing this package?

matheuscscp (approved):
LGTM! :rocket:

internal/test/external_artifact_test.go
stefanprodan: You can't run an actual controller in the main tests, everything with break due the mix of mock client and etcd. What will happen is that the controller will reconcile all the objects in all the tests.

internal/controller/helmrelease_controller.go
stefanprodan: Two reasons:
- ExternalArtifact does not implement the `fluxcd/pkg/runtime/conditions.Getter` interface 
- Reconcilers shouldn't care if the source is ready or not, they should only care if an external artifact has the tarball with the chart available for download

stefanprodan: I made ExternalArtifact implement `Getter` but the 2nd point still stands, unlike the native sources that HC may generate, the external artifact can have a valid chart even when not ready, KC does this for all sources.

PR #1304, opened 2025-09-22
Update source-controller API to v1.7.0

stefanprodan:
ref: https://github.com/fluxcd/source-controller/releases/tag/v1.7.0


matheuscscp (approved):
LGTM! :rocket:

Issue #896, opened 2024-02-14
Drift mode should detect extra properties


darkweaver87:
Hello,

When drift mode is set either to `warn` or `enabled`, it should detect extra map keys and extra list objects.
For instance if an `HelmRelease` installs a `Deployment` with some environment variables set, it should detect this deployment has some manually added extra environment variables for instance and with correction enabled, remove those.

Thanks,

Rémi

hoerup:
(using flux 2.2.3 )

Just modified one of my Deployments which is installed via a HelmRelease and with driftmode enabled and added a dns override
`spec:
  template:
    spec:
      hostAliases:
      - hostnames:
        - somehostname
        ip: 127.0.0.1
        `
and this is not detected by flux

ViacheslavKudinov:
Hi,
we see same issue.
Controller sees only change the values which exist on deployment, but extra env variables are not detected as drift.
```
flux check
► checking prerequisites
✔ Kubernetes 1.32.0 >=1.30.0-0
► checking version in cluster
✔ distribution: flux-v2.5.1
✔ bootstrapped: true
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v1.2.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v1.5.1
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v1.5.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v1.5.0
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta3
✔ buckets.source.toolkit.fluxcd.io/v1
✔ gitrepositories.source.toolkit.fluxcd.io/v1
✔ helmcharts.source.toolkit.fluxcd.io/v1
✔ helmreleases.helm.toolkit.fluxcd.io/v2
✔ helmrepositories.source.toolkit.fluxcd.io/v1
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1
✔ ocirepositories.source.toolkit.fluxcd.io/v1beta2
✔ providers.notification.toolkit.fluxcd.io/v1beta3
✔ receivers.notification.toolkit.fluxcd.io/v1
✔ all checks passed
```

PR #1365, opened 2025-11-27
Add --override-manager flag for server-side apply drift detection

yozel:
This flag allows specifying field managers whose ownership should be transferred to the helm-controller before performing drift detection. When a disallowed field manager is detected on a managed resource, its field ownership is replaced with the helm-controller's field owner, enabling proper drift detection for fields that were previously modified by other controllers or tools (e.g., kubectl).

This feature mirrors the --override-manager flag available in FluxCD's kustomize-controller, providing consistent behavior across Flux components for managing server-side apply field ownership conflicts.

Since it doesn't change the default behavior, the implementation doesn't introduce any breaking changes.

Fix: https://github.com/fluxcd/helm-controller/issues/896

### DIFF
--- a/internal/action/diff.go
+++ b/internal/action/diff.go
@@ -27,6 +27,8 @@ import (
 	helmaction "helm.sh/helm/v3/pkg/action"
 	helmrelease "helm.sh/helm/v3/pkg/release"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	apierrutil "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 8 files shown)


yozel:
@stefanprodan just signed-off the commit 👍

yozel:
It passes all the tests and checks. @matheuscscp could you take look when you have time? If you have any comments or suggestions, I can work on them immediately.

internal/action/diff.go
```diff
@@ -278,3 +293,39 @@ func (s sortableDiffs) Less(i, j int) bool {
 
 	return iDiff.GetName() < jDiff.GetName()
 }
+
+func replaceDisallowedFieldManagers(ctx context.Context, c client.Client, fieldOwner string, disallowedFieldManagers []string, obj *unstructured.Unstructured) error {
+	existingObj := &unstructured.Unstructured{}
+	existingObj.SetGroupVersionKind(obj.GroupVersionKind())
+	if err := c.Get(ctx, client.ObjectKeyFromObject(obj), existingObj); client.IgnoreNotFound(err) != nil {
```
matheuscscp: Will the patch below work if the object does not exist?

yozel: It is supposed to work since it will result with empty `[]JSONPatch`, but it is actually better to make it explicit and return early in case if the object does not exist. I've just changed it.

stefanprodan:
Please squash all commits and force push

Issue #1241, opened 2025-06-18
Drift detection in warn mode - having status field or metrics to have alerting


sebglon:
Hi,
We need to monitor and be alerted when the helm-controller detect a drift.
can we have a status field or a metrics to implement alerting?

sebglon:
We use unified monitoring and alerting with prometheus.
For that we need to have metrics to monitor the drift resolution.

stefanprodan:
> We use unified monitoring and alerting with prometheus.

So you can use the Prometheus Alert Manager integration https://fluxcd.io/flux/components/notification/providers/#prometheus-alertmanager

matheuscscp:
You can plot alert graphs in Grafana, have you tried setting up Alertmanager as a data source?

https://grafana.com/docs/grafana/latest/datasources/alertmanager/

sebglon:
If we have a status.driftDetected fields on the HelmRelease object with the reason, we can easily generate metrics and take actions to fix the drift.
This can be the same for `warn` and `enabled`. 
We have to identify helmReleae reconciliation loop with driftDetection.
And we can't enable debug logs in production due to data volume generated

matheuscscp:
You can have one Provider+Alert per namespace

sebglon:
Why not for the Flux Operator but this not solve the Metrics and the clear view of the drift without using debug logs.

We need to add a HelmRelease status condition with the drift details. this help each teams and tenants to debug and monitor drift without using the flux debug logs...

PR #1367, opened 2025-12-04
Add `Drifted` condition to HelmRelease

yozel:
This PR Introduces a new condition type "Drifted" to improve observability of Helm release drift detection.

Fix: #1241

### DIFF
--- a/internal/reconcile/atomic_release.go
+++ b/internal/reconcile/atomic_release.go
@@ -381,6 +381,10 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state 
 			}
 		}
 
+		if req.Object.GetDriftDetection().MustDetectChanges() {
+			conditions.MarkFalse(req.Object, v2.DriftedCondition, v2.NoDriftDetectedReason, "No drift detected against the cluster state")
+		}
+
 		return nil, nil
 	case ReleaseStatusLocked:
 		log.Info(msgWithReason("release locked", state.Reason))
@@ -436,10 +440,10 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state 
 			}
 		}
 
-		r.eventRecorder.Eventf(req.Object, corev1.EventTypeWarning, "DriftDetected",
-			"Cluster state of release %s has drifted from the desired state:\n%s",
-			req.Object.Status.History.Latest().FullReleaseName(), diff.SummarizeDiffSet(state.Diff),
-		)
+		msg := fmt.Sprintf("Cluster state of release %s has drifted from the desired state:\n%s",
+			req.Object.Status.History.Latest().FullReleaseName(), diff.SummarizeDiffSet(state.Diff))
+		r.eventRecorder.Eventf(req.Object, corev1.EventTypeWarning, v2.DriftDetectedReason, msg)
+		conditions.MarkTrue(req.Object, v2.DriftedCondition, v2.DriftDetectedReason, "%s", msg)
 
 		if req.Object.GetDriftDetection().GetMode() == v2.DriftDetectionEnabled {
 			return NewCorrectClusterDrift(r.configFactory, r.eventRecorder, state.Diff, kube.ManagedFieldsManager), nil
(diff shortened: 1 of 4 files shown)


internal/reconcile/atomic_release.go
```diff
@@ -385,6 +385,14 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state
 			conditions.MarkTrue(req.Object, v2.ReleasedCondition, v2.UpgradeSucceededReason, "%s", msg)
 		}
 
+		if req.Object.GetDriftDetection().MustDetectChanges() {
+			conditions.Set(req.Object, &metav1.Condition{
```
matheuscscp: Please use `conditions.MarkFalse()`. Also, `Message` is a required field. Let's use:

`No drift detected against the cluster state.`

internal/reconcile/atomic_release.go
```diff
@@ -440,10 +448,15 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state
...
 		}
 
-		r.eventRecorder.Eventf(req.Object, corev1.EventTypeWarning, "DriftDetected",
-			"Cluster state of release %s has drifted from the desired state:\n%s",
-			req.Object.Status.History.Latest().FullReleaseName(), diff.SummarizeDiffSet(state.Diff),
-		)
+		msg := fmt.Sprintf("Cluster state of release %s has drifted from the desired state:\n%s",
+			req.Object.Status.History.Latest().FullReleaseName(), diff.SummarizeDiffSet(state.Diff))
+		r.eventRecorder.Eventf(req.Object, corev1.EventTypeWarning, v2.DriftDetectedReason, msg)
+		conditions.Set(req.Object, &metav1.Condition{
```
matheuscscp: Please use `conditions.MarkTrue(..., "%s", msg)`. Also, I think maybe it only makes sense to mark this condition as true if the HelmRelease is not configured to correct the drift.

yozel: I'll convert it to `conditions.MarkTrue(..., "%s", msg)`.

About your second point, it still might make sense to have this condition true just before the correction so if the correction has an error, the drift status is still observable. What do you think?

matheuscscp: Yeah, makes sense!

internal/reconcile/atomic_release.go
yozel: Fixed

internal/reconcile/atomic_release.go
yozel: Changed it to `conditions.MarkTrue`

matheuscscp:
It would be nice to add a test for when the condition is false

matheuscscp (approved):
LGTM! :rocket:

internal/reconcile/atomic_release.go
```diff
@@ -385,6 +385,10 @@ func (r *AtomicRelease) actionForState(ctx context.Context, req *Request, state
 			conditions.MarkTrue(req.Object, v2.ReleasedCondition, v2.UpgradeSucceededReason, "%s", msg)
 		}
 
+		if req.Object.GetDriftDetection().MustDetectChanges() {
+			conditions.MarkFalse(req.Object, v2.DriftedCondition, v2.NoDriftDetectedReason, "No drift detected against the cluster state")
+		}
+
```
stefanprodan, line 387: Shouldn't we remove this condition if drift is not enabled in spec?

matheuscscp: Maybe put an `else` to remove the condition if `MustDetectChanges()` returns false?

yozel: @stefanprodan good point.

@matheuscscp then it would not be removed in case release state is not `InSync`. I can remove the condition in the beginning of `AtomicRelease.actionForState` [here](https://github.com/yozel/helm-controller/blob/0d71cdf461b475babedca0880e6686557dd0fcd9/internal/reconcile/atomic_release.go#L337) or even in `AtomicRelease.Reconcile` [here](https://github.com/yozel/helm-controller/blob/0d71cdf461b475babedca0880e6686557dd0fcd9/internal/reconcile/atomic_release.go#L191), so we can be sure the condition is removed in case drift detection disabled later on

matheuscscp: Works for me

yozel: Done

Issue #1381, opened 2026-01-12
Add support for server-side apply


matheuscscp:
Depends on: #1300

Helm 4 has introduced support for Kubernetes Server-Side Apply. Now we can introduce it in helm-controller.

For this, we essentially need to pass any inputs related to server-side apply that the Helm SDK allows in the `internal/action` package and add some (preferably e2e) tests for it.

From what I can see of the SDK API, we need the following new fields in HelmRelease:

```yaml
spec:
  install:
    serverSideApply: true # or false. boolean. defaults to true
  upgrade:
    serverSideApply: "true" # or "false", or "auto". string. defaults to "auto"
  rollback:
    serverSideApply: "true" # or "false", or "auto". string. defaults to "auto"
```

These fields should be defined in the respective Go structs with a pointer, so `false` can be distinguished from using the default. This is important because the default changes depending on the feature gate `UseHelm3Defaults`.

stefanprodan:
I suggest we don't expose `forceConflicts` and we set it to `true`, this is how kustomize-controller does SSA and IMO helm-controller should match the behaviour. 

PS. If we allow `forceConflicts: false` then our drift detection feature will break.

matheuscscp:
Agreed! I edited the description.

matheuscscp:
Agreed. It will also be hard to maintain and explain the Flux behavior in the future if the defaults diverge from Helm.

PR #1384, opened 2026-01-15
Add ServerSideApply field to HelmRelease API

cappyzawa:
Resolves: #1381

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -63,6 +63,80 @@ jobs:
           kubectl -n helm-system wait helmreleases/podinfo-git --for=condition=ready --timeout=4m
           kubectl -n helm-system wait helmreleases/podinfo-oci --for=condition=ready --timeout=4m
           kubectl -n helm-system delete -f config/testdata/podinfo
+      - name: Run server-side apply test
+        run: |
+          test_name=server-side-apply
+          kubectl -n helm-system apply -f config/testdata/$test_name/install.yaml
+          kubectl -n helm-system wait helmreleases/$test_name --for=condition=ready --timeout=4m
+
+          # Verify the release is deployed with SSA.
+          APPLY_METHOD=$(kubectl -n helm-system get secret sh.helm.release.v1.$test_name.v1 -o jsonpath='{.data.release}' | base64 -d | base64 -d | gunzip | jq -r '.apply_method')
+          if [ "$APPLY_METHOD" != "ssa" ]; then
+            echo -e "Unexpected apply method: $APPLY_METHOD (expected: ssa)"
+            exit 1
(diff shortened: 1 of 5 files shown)


api/v2/helmrelease_types.go
```diff
@@ -763,6 +768,14 @@ type Upgrade struct {
...
 	// +optional
 	CRDs CRDsPolicy `json:"crds,omitempty"`
+
+	// ServerSideApply enables server-side apply for resources during upgrade.
+	// Can be "true", "false", or "auto".
+	// When "auto", server-side apply usage will be based on the release's previous usage.
+	// Defaults to "auto".
+	// +kubebuilder:validation:Enum={"true","false","auto"}
+	// +optional
+	ServerSideApply *string `json:"serverSideApply,omitempty"`
```
matheuscscp:
```suggestion
	ServerSideApply string `json:"serverSideApply,omitempty"`
```

Since we don't use the empty string for anything, the string fields from upgrade and rollback fields don't need to be pointers :+1:

api/v2/helmrelease_types.go
```diff
@@ -1021,6 +1034,14 @@ type Rollback struct {
 	// rollback action when it fails.
 	// +optional
 	CleanupOnFail bool `json:"cleanupOnFail,omitempty"`
+
+	// ServerSideApply enables server-side apply for resources during rollback.
+	// Can be "true", "false", or "auto".
```
stefanprodan: We've been avoiding booleans as strings in HelmRelease API because they get converted automatically by KC at apply and will result in an error.

I suggest using: `enabled`, `disabled`, `auto`.

matheuscscp: Fair enough. They still need to be converted to `"true"` and `"false"`, it's what the Helm SDK expects

api/v2/helmrelease_types.go
cappyzawa: That’s true!
I'll fix it.

api/v2/helmrelease_types.go
cappyzawa: Got it! Makes sense to avoid the YAML boolean auto-conversion issue. I'll update to use enabled/disabled/auto instead.

cappyzawa: fixed via [51a7c21](https://github.com/fluxcd/helm-controller/pull/1384/commits/51a7c2130c5ca04cccc69d2d3c30e4a9c682d29e) and [848a73d](https://github.com/fluxcd/helm-controller/pull/1384/commits/848a73d5fb163f969b7a7f764dac2b19016a8eaf).

cappyzawa:
Squashed the fixups, ready for merge 🙏

Issue #1352, opened 2025-11-18
Keep track of managed objects in `.status.inventory`


stefanprodan:
Similar to Flux Kustomization, the HelmRelease should expose the managed objects in `.status.inventory`. This would improve the observability of Flux as right now users need to query the Helm storage, extract the tar.gz from the Secret value and iterate over the objects to list them. One major pain point with Helm storage lookup is that the managed objects don't have the namespace set, CRDs are missing, etc. The helm-controller knows which objects it applies, so we can generate an accurate inventory.

matheuscscp:
The Flux CLI and other tools that traverse the resource tree can also benefit, no more .tar.gz processing anywhere.

stealthybox:
would this also track resources such as upgrade hooks and tests in the same inventory list?
Just noting this should specifically use the output of post-renderers.

matheuscscp:
This would reflect the rendered manifests you can find by opening up the helm storage secret

matheuscscp:
> CRDs are not in the secret, but since we manage them unlike Helm, they should be in the inventory.

Yep, with addition of the CRDs

Ashish7812:
@stefanprodan and @matheuscscp Can we also export the status field of some selected resources of the chart to the helm release status field? 

A user can select which k8 resource status field to export via some annotation. This way if there is dependency between two helm releases then the dependent helm release can read the output and can use it in its own helm release as input value.

ex: Lets say in the helm release we have the following resource along with others

```yaml

apiVersion: ec2.services.k8s.aws/v1alpha1
kind: VPC
metadata:
  name: tutorial-vpc-{{ .Release.Name }}
  annotations:
    output_vpcid: "$.status.vpcID" # Output this key in the status field
spec:
  cidrBlocks:
  - 10.0.0.0/16
  enableDNSSupport: true
  enableDNSHostnames: true
  tags:
    - key: name
      value: vpc-tutorial-{{ .Release.Name }}
```

The helm release can read the status field of the VPC resource and output the selected key.

PR #1385, opened 2026-01-16
Add `.status.inventory` to track managed objects

cappyzawa:
Resolves: #1352


internal/reconcile/atomic_release.go
```diff
@@ -234,6 +236,11 @@ func (r *AtomicRelease) Reconcile(ctx context.Context, req *Request) error {
 						// Update the common-metadata digest if common-metadata exist.
 						req.Object.Status.ObservedCommonMetadataDigest = postrender.CommonMetadataDigest(digest.Canonical, req.Object.Spec.CommonMetadata).String()
 					}
+
+					if err := r.buildInventory(ctx, req); err != nil {
+						// Inventory build failure should not fail the reconciliation.
```
cappyzawa: #1352 does not specify error handling behavior for inventory build failures.

I chose to log the error and continue reconciliation because:
- The inventory is supplementary metadata, not critical to the release operation
- A failed inventory build shouldn't prevent an otherwise successful release

However, this means `.status.inventory` could be empty/stale if　`buildInventory` fails.

Should we instead fail the reconciliation, or is this approach acceptable?

cappyzawa: > [!NOTE]
> In kustomize-controller, inventory build failure causes reconciliation to fail ([kustomization_controller.go](https://github.com/fluxcd/kustomize-controller/blob/b531fe14ccbbe7c4a5857c447838f653b477f618/internal/controller/kustomization_controller.go#L481-L488)).
>
> However, the context differs:
> - kustomize-controller: Inventory is built from SSA's `ChangeSet` (the apply result itself), so if apply succeeds, inventory build should also succeed.
> - helm-controller: Inventory is built separately by parsing `release.Manifest` from Helm storage and querying the RESTMapper for namespace scope. This can fail independently of a successful release.

matheuscscp: I think you raise excellent points here and I agree with them. Since this state is superfluous for the HelmRelease state machine, I think we should not mark the object as failed in this case. @hiddeco can confirm if this is safe.

I'm wondering, however, if reporting a partial inventory (i.e. by using at least the objects that were successfully found and returning a list of errors for the objects where we had some error) can be beneficial. I can imagine that a partial inventory can be confusing for users, but since this information is just for observability I can see it being useful. So I'm really not sure about this.

Thoughts anybody? @hiddeco @stefanprodan

internal/reconcile/release.go
```diff
@@ -126,6 +133,46 @@ func observeRelease(observed observedReleases) storage.ObserveFunc {
...
+func observeInventory(obj *v2.HelmRelease, chart *helmchart.Chart, getter genericclioptions.RESTClientGetter, recorder record.EventRecorder) storage.ObserveFunc {
+	return func(rls helmrelease.Releaser) {
+		rlsTyped, ok := rls.(*helmreleasev1.Release)
+		if !ok {
+			return
+		}
+
+		restCfg, err := getter.ToRESTConfig()
+		if err != nil {
+			recorder.Eventf(obj, corev1.EventTypeWarning, "InventoryBuildFailed",
```
matheuscscp: Maybe add a `const` for the reason on the beginning of `func(rls helmrelease.Releaser) {`.

internal/reconcile/release.go
```diff
@@ -126,6 +133,46 @@ func observeRelease(observed observedReleases) storage.ObserveFunc {
...
+				"failed to build inventory for release %s: %s", obj.GetReleaseName(), err.Error())
```
matheuscscp:
```suggestion
				"failed to build inventory for %s/%s: %s", obj.GetNamespace(), obj.GetName(), err.Error())
```

internal/inventory/inventory.go
```diff
@@ -0,0 +1,127 @@
...
+
+// AddManifest parses the manifest, complements namespaces, and adds the objects to the inventory.
+func AddManifest(inv *v2.ResourceInventory, manifest string, releaseNamespace string, c client.Client) error {
+	objects, err := parseManifest(manifest)
+	if err != nil {
+		return fmt.Errorf("failed to parse manifest: %w", err)
+	}
+
+	// NOTE: Helm hooks are ephemeral resources that are deleted after execution,
+	// so they should not be tracked in the inventory.
```
hiddeco: This technically depends on the type of policy that is being set: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies

But I agree on them not being tracked due to Helm considering them to not be part of the release itself: https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

internal/reconcile/release.go
```diff
@@ -126,6 +133,46 @@ func observeRelease(observed observedReleases) storage.ObserveFunc {
...
+			return
+		}
+		c, err := client.New(restCfg, client.Options{})
```
hiddeco, line 153: Do we have to construct a new client every time, can't it be injected?

cappyzawa: Helm doesn't set .metadata.namespace on namespaced objects in the manifest, so I need to complement it. I referenced action.Diff() for the namespace completion pattern (using apiutil.IsObjectNamespaced()), which creates a client per invocation. I followed that established pattern here.

internal/reconcile/release.go
cappyzawa: Added InventoryBuildFailedReason to api/v2/inventory_types.go following the pattern in condition_types.go ([0c672c1](https://github.com/fluxcd/helm-controller/pull/1385/commits/0c672c19c0363b73f8d44c1e605cfd56cc576d3b))

internal/reconcile/release.go
matheuscscp: I don't think creating a client makes any calls, so it's a cheap operation (pls correct me if I'm wrong). I don't mind addressing this optimization separately from this PR, looks like there are more places prior to this PR where we could make the same improvement

stefanprodan: We can leave this for another PR, I don't see much of a perf impact recreating the client here to resolve the namespaced objects.

internal/reconcile/atomic_release.go
stefanprodan: I would log the RESTMapper query errors and skip setting the namespace. If the storage parser fail, I would fail the HR as well, this means a corrupted storage so nothing will work going forward.

internal/inventory/inventory.go
matheuscscp: I think we could just remove this comment since there's no mention to hooks anywhere around this code anyway

internal/reconcile/atomic_release.go
cappyzawa: Thanks for your advice.
Fixed via [97fd9aa](https://github.com/fluxcd/helm-controller/pull/1385/commits/97fd9aac750d7c50c7dce85ea4ad8eb5ef25d98f)

internal/inventory/inventory.go
cappyzawa: make sense.
Fixed via [1905a9a](https://github.com/fluxcd/helm-controller/pull/1385/commits/1905a9a39f614cc18c4072ffe2edb8861aaee5e4)

docs/spec/v2/helmreleases.md
```diff
@@ -1884,6 +1884,29 @@ status:
...
+### Inventory
+
+The HelmRelease reports the list of Kubernetes resource objects that have been
+applied by the Helm release in `.status.inventory`. This can be used to
+identify which objects are managed by the HelmRelease. The inventory records
+are in the format `<namespace>_<name>_<group>_<kind>`.
+
+The inventory includes all resources from the rendered manifests, as well as
+CRDs from the chart's `crds/` directory. Helm hooks are not included in the
+inventory, as they are not considered part of the release by Helm.
```
matheuscscp, line 1896: Best place to mention hooks :+1:

docs/spec/v2/helmreleases.md
```diff
@@ -1884,6 +1884,29 @@ status:
...
+
+```console
+Status:
```
matheuscscp: Looks like in the history example that was added recently we are using the YAML format from `kubectl get hr <name> -o yaml`

internal/inventory/inventory.go
```diff
@@ -0,0 +1,124 @@
...
+
+func parseManifest(manifest string) ([]*unstructured.Unstructured, error) {
+	return ssautil.ReadObjects(strings.NewReader(manifest))
+}
+
+// setNamespaces complements namespace for namespaced objects that don't have one set.
+// This is necessary because Helm manifests don't include namespace for namespaced resources.
+func setNamespaces(objects []*unstructured.Unstructured, releaseNamespace string, c client.Client) []string {
+	var warnings []string
+	isNamespacedGVK := map[string]bool{}
```
matheuscscp:
```suggestion
	isNamespacedGVK := map[schema.GroupVersionKind]bool{}
```

internal/inventory/inventory.go
```diff
@@ -0,0 +1,124 @@
...
+
+	for _, obj := range objects {
+		if obj.GetNamespace() != "" {
+			continue
+		}
+
+		objGVK := obj.GetObjectKind().GroupVersionKind().String()
```
matheuscscp:
```suggestion
		objGVK := obj.GetObjectKind().GroupVersionKind()
```

internal/reconcile/release.go
```diff
@@ -126,6 +133,50 @@ func observeRelease(observed observedReleases) storage.ObserveFunc {
...
+		}
+
+		inv := inventory.New()
+		warnings, err := inventory.AddManifest(inv, rlsTyped.Manifest, rlsTyped.Namespace, c)
+		if err != nil {
+			recorder.Eventf(obj, corev1.EventTypeWarning, v2.InventoryBuildFailedReason,
+				"failed to build inventory for %s/%s: %s", obj.GetNamespace(), obj.GetName(), err.Error())
+			return
+		}
+		for _, w := range warnings {
```
matheuscscp: I think here we need to concat the warnings into a single message to avoid sending too many events for a single storage observation, these events land  in notification-controller as well (hence in Slack :grin:)

internal/reconcile/atomic_release.go
matheuscscp: In your commit I see you implemented a middle-ground solution by only sending error events in case parsing the manifest fails. I agree with this approach... I think we don't need to introduce a new failure mode here when a reconciliation would other be successful, this sounds risky to me. If the storage is corrupted, I'd really like to believe that something is in place today for users to know. On top of that, we are now also sending an error event to alert users. I think this is the perfect middle-ground solution here.

internal/reconcile/release.go
stefanprodan: Yes we need to contact, as this will create too much spam if the API server goes down the moment we resolve the namespace for all objects.

internal/reconcile/atomic_release.go
matheuscscp: Going to mark this as as resolved for now, if we still want to introduce this new failure mode we can do it later. Skipping it now is to avoid grave consequences.

internal/reconcile/release.go
cappyzawa: Done. Concatenated warnings with `strings.Join(warnings, "; ")`. https://github.com/fluxcd/helm-controller/pull/1385/commits/9fee3db096ea28acb378f70afa6e0a74205c7c2b

internal/inventory/inventory.go
cappyzawa: Done(https://github.com/fluxcd/helm-controller/pull/1385/commits/9fee3db096ea28acb378f70afa6e0a74205c7c2b)

internal/inventory/inventory.go
cappyzawa: Done(https://github.com/fluxcd/helm-controller/pull/1385/commits/9fee3db096ea28acb378f70afa6e0a74205c7c2b)

docs/spec/v2/helmreleases.md
cappyzawa: Done(https://github.com/fluxcd/helm-controller/pull/1385/commits/9fee3db096ea28acb378f70afa6e0a74205c7c2b)

Issue #1382, opened 2026-01-12
Add support for custom health checks via CEL expressions


matheuscscp:
Depends on: #1300

Helm 4 has introduced support for passing custom kstatus readers. We can now pass our custom kstatus readers built from CEL expressions like in kustomize-controller.

References:
- `pkg` library support: https://github.com/fluxcd/pkg/pull/1068 (allows building a single `engine.StatusReader` from `[]kustomize.CustomHealthCheck`)
- kustomize-controller implementation: https://github.com/fluxcd/kustomize-controller/pull/1344
- Helm support: https://github.com/helm/helm/pull/31706 (allows passing custom `engine.StatusReader` to the Actions SDK)

Note: The helm-controller `api/v2` package already uses `fluxcd/pkg/apis/kustomize`. We can safely use the type `[]kustomize.CustomHealthCheck` there.

## Important note about this feature

It will only be available for kstatus health checks implemented in Helm 4, i.e. the `waiter` `WaitStrategy`. Users that opt into the `legacy` `WaitStrategy` will not be able to use this feature.

PR #1389, opened 2026-01-23
Add support for custom health checks via CEL expressions

matheuscscp:
Closes: #1382

Implements [RFC-0009](https://github.com/fluxcd/flux2/tree/main/rfcs/0009-custom-health-checks) for helm-controller.


docs/spec/v2/helmreleases.md
```diff
@@ -942,6 +942,88 @@ spec:
...
+  chartRef:
+    kind: OCIRepository
+    name: sealed-secrets-chart
+  values:
+    replicaCount: 2
+  healthCheckExprs:
+    - apiVersion: bitnami.com/v1alpha1
+      kind: SealedSecret
+      failed: status.conditions.filter(e, e.type == 'Synced').all(e, e.status == 'False')
+      current: status.conditions.filter(e, e.type == 'Synced').all(e, e.status == 'True')
```
Ashish7812, line 1010: @matheuscscp  if there are multiple **SealedSecret** instances in the chart, will the helm release wait for all to reach the Synced status?

matheuscscp: This field is per type, not per object.

Ashish7812: @matheuscscp  > This field is per type, not per object.

Yes, based on the declaration I got it but I am asking if the chart contains multiple SealedSecret will helm release wait for the SealedSecret type CR to be in successful terminal status?

Out of two SealedSecret CR present in the chart if one is taking more time to reach the terminal status, then I believe it will wait for that also. is my understanding correct?

matheuscscp: Flux always waits for all resources to become ready if no resources have failed so far. It fails fast if at least one resource fails.

Issue #1380, opened 2026-01-12
Reduce the mean time to recovery (MTTR) in case of failed deployments


matheuscscp:
Depends on: #1391

Issue for tracking the implementation of https://github.com/fluxcd/flux2/issues/5584 in helm-controller.

References:
- `pkg` library support: https://github.com/fluxcd/pkg/pull/1042 (allows fetching a `context.Context` that will be canceled on new enqueue events)
- kustomize-controller implementation: https://github.com/fluxcd/kustomize-controller/pull/1536
- Helm support: https://github.com/helm/helm/pull/31735 (allows passing a custom `context.Context` for health checks)

## Important note about this feature

It will only be available for kstatus health checks implemented in Helm 4, i.e. the `waiter` `WaitStrategy`. Users that opt into the `legacy` `WaitStrategy` will not be able to use this feature.

PR #1392, opened 2026-01-27
Reduce the mean time to recovery (MTTR) in case of failed deployments

matheuscscp:
Depends on: https://github.com/fluxcd/pkg/pull/1082

Closes: #1380
Closes: https://github.com/fluxcd/flux2/issues/5584

### DIFF
--- a/internal/controller/helmrelease_controller.go
+++ b/internal/controller/helmrelease_controller.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	"github.com/Masterminds/semver/v3"
+	"github.com/fluxcd/cli-utils/pkg/kstatus/polling"
+	"github.com/fluxcd/cli-utils/pkg/kstatus/polling/clusterreader"
 	"github.com/fluxcd/cli-utils/pkg/kstatus/polling/engine"
 	celtypes "github.com/google/cel-go/common/types"
 	chart "helm.sh/helm/v4/pkg/chart/v2"
@@ -58,6 +60,7 @@ import (
 	"github.com/fluxcd/pkg/runtime/logger"
 	"github.com/fluxcd/pkg/runtime/object"
 	"github.com/fluxcd/pkg/runtime/patch"
+	"github.com/fluxcd/pkg/ssa"
 
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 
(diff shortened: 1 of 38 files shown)


internal/controller/helmrelease_controller.go
```diff
@@ -376,7 +377,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 		// If this fails, the controller will fall back to performing an upgrade
 		// to settle on the desired state.
 		// TODO(hidde): remove this in a future release.
-		if err := r.adoptLegacyRelease(ctx, getter, statusReader, obj); err != nil {
+		if err := r.adoptLegacyRelease(ctx, getter, obj); err != nil {
```
matheuscscp: Removing status reader from here since `adoptLegacyRelease` performs no wait operations

main.go
```diff
@@ -288,7 +294,7 @@ func main() {
 
 	if watchNamespace != "" {
 		mgrConfig.Cache.DefaultNamespaces = map[string]ctrlcache.Config{
-			watchNamespace: ctrlcache.Config{},
```
matheuscscp, line 296: Fixing linting error

internal/controller/helmrelease_controller.go
matheuscscp: @stefanprodan 

>  // TODO(hidde): remove this in a future release.

Has this future arrived?

matheuscscp: Since v2beta1 was removed: https://github.com/fluxcd/helm-controller/pull/1280

stefanprodan: Yeap we can delete this now

internal/action/job_status_reader.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/fluxcd/cli-utils/pkg/kstatus/polling/statusreaders"
+	"github.com/fluxcd/cli-utils/pkg/kstatus/status"
+	"github.com/fluxcd/cli-utils/pkg/object"
+)
+
+type customJobStatusReader struct {
+	genericStatusReader engine.StatusReader
+}
+
+func NewCustomJobStatusReader(mapper meta.RESTMapper) engine.StatusReader {
```
stefanprodan: Replace this with https://github.com/fluxcd/pkg/blob/main/runtime/statusreaders/job.go

matheuscscp: Fixed, thanks!

PR #1398, opened 2026-01-30
Introduce description field to .status.history

matheuscscp:
In Helm's upcoming Release API v2 (coming for supporting Chart API v3), the `status` field may get a fine-grained value:

https://github.com/banjoh/helm/pull/1

`deployed` -> `installed`, `upgraded`, `rolledback`

Currently, `status` already has `uninstalled`.


matheuscscp:
Better track the action instead

PR #1399, opened 2026-02-01
Track action in snapshots

matheuscscp:
Introduce `.action` field for entries in `.status.history`.

### DIFF
--- a/internal/release/observation.go
+++ b/internal/release/observation.go
@@ -67,6 +67,10 @@ type Observation struct {
 	Version int `json:"version"`
 	// Info provides information about the release.
 	Info helmrelease.Info `json:"info"`
+	// Action is the action that resulted in this observation.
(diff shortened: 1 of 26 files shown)


internal/release/observation.go
```diff
@@ -67,6 +67,10 @@ type Observation struct {
 	Version int `json:"version"`
 	// Info provides information about the release.
 	Info helmrelease.Info `json:"info"`
+	// Action is the action that resulted in this observation.
+	// It's not serialized to JSON because it's not needed for
+	// the digest. The action is only tracked for visibility.
+	Action v2.ReleaseAction `json:"-"`
```
matheuscscp, line 73: Since this field is purely a visibility feature I'm avoiding adding it to the digest calculation for now, a ton of tests failed adding it to the digest.

PR #1407, opened 2026-02-13
Add `DirectSourceFetch` feature gate to bypass cache for source objects

dipti-pai:
This feature gate enables fetching source objects directly from the API server using APIReader, bypassing the controller's cache. This can be useful when immediate consistency is required for source object reads.

When enabled via --feature-gates=DirectSourceFetch=true:

Source objects are fetched using r.APIReader instead of r.Client A log message is emitted at startup indicating the feature is active Changes:

Add DirectSourceFetch field to HelmReleaseReconciler struct Update getSource() to use APIReader when feature is enabled Register feature gate with default value false (opt-in) Add unit tests
Update pkg/runtime dependency to v0.100.1

Part of : https://github.com/fluxcd/kustomize-controller/issues/1583

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,8 +25,8 @@ require (
 	github.com/fluxcd/pkg/auth v0.38.2
 	github.com/fluxcd/pkg/cache v0.13.0
 	github.com/fluxcd/pkg/chartutil v1.22.0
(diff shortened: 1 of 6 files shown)


matheuscscp (approved):
LGTM! 🚀

PR #1412, opened 2026-02-18
Fix postRenderers not causing new upgrade when applied during ongoing upgrade

matheuscscp:
Fixes a race condition where changing `postRenderers` or `commonMetadata` while the controller is performing an upgrade does not trigger a second upgrade to apply the new configuration.

**Root cause:** The generation guard in `DetermineReleaseState` compared `req.Object.Generation` against the Ready condition's `ObservedGeneration` to decide whether to check for digest changes. However, `patchStatusConditions()` re-fetches the object from the API server to resolve conflicts, and `conditions.Set()` always overwrites `ObservedGeneration` with `to.GetGeneration()` — which at that point reflects the latest generation (including the mid-reconciliation spec change). This causes the guard to see the generation as already processed, skipping the digest comparison entirely on the next reconciliation.

**Fix:**
- Use the top-level `status.observedGeneration` for the generation guard instead of the Ready condition's `ObservedGeneration`. The top-level field is set by the outer defer using the in-memory generation snapshot and is not affected by the patch helper's conflict resolution.
- Move the `ObservedPostRenderersDigest` and `ObservedCommonMetadataDigest` updates into `summarize()`, which is called after every action (install, upgrade, test, rollback, etc.), ensuring the digests are always kept in sync.

### DIFF
--- a/internal/reconcile/atomic_release.go
+++ b/internal/reconcile/atomic_release.go
@@ -38,9 +38,7 @@ import (
 	v2 "github.com/fluxcd/helm-controller/api/v2"
 	"github.com/fluxcd/helm-controller/internal/action"
 	"github.com/fluxcd/helm-controller/internal/diff"
-	"github.com/fluxcd/helm-controller/internal/digest"
 	interrors "github.com/fluxcd/helm-controller/internal/errors"
-	"github.com/fluxcd/helm-controller/internal/postrender"
 )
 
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

stealthybox:
both your script and my tests are validated as failing on main and passing on this PR, now that I'm building for arm64

awesome fix on this
LGTM

Issue #1429, opened 2026-03-03
HelmRelease stuck as not ready (CancelHealthCheckOnNewRevision)


artem-nefedov:
## Description

Using flux 2.8.1 (helm-controller v1.5.1) with `--feature-gates=CancelHealthCheckOnNewRevision=true`, I would sometimes get a problem where HelmRelease would fail to install, seemingly due to healthchecks getting cancelled for no reason (`New reconciliation triggered by HelmChart ...`), and then would get stuck as not ready. Its status also shows that it tries to upgrade, despite no changes being made to either source or HelmRelease itself since their creation, and the cluster being brand new. I don't know the exact reproduction steps, and I only get this problem occasionally, but it happens often enough to be the blocker. If `CancelHealthCheckOnNewRevision` feature flag is removed, the problem is not reproduced.

## Involved objects, as seen in the cluster after the problem occurred

* HelmRelease (with managed fields, to demonstrate that nothing except helm-controller updated it after creation)

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  creationTimestamp: "2026-03-03T15:38:10Z"
  finalizers:
  - finalizers.fluxcd.io
  generation: 1
  name: "<redacted>"
  namespace: "<redacted>"
  resourceVersion: "<redacted>"
  uid: "<redacted>"
  managedFields:
  - apiVersion: helm.toolkit.fluxcd.io/v2
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:chart:
          f:spec:
            f:chart: {}
            f:reconcileStrategy: {}
            f:sourceRef:
              f:kind: {}
              f:name: {}
              f:namespace: {}
            f:version: {}
        f:interval: {}
        f:timeout: {}
        f:values: {} # <collapsed>
    manager: kubectl
    operation: Apply
    time: "2026-03-03T15:30:35Z"
  - apiVersion: helm.toolkit.fluxcd.io/v2
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .: {}
          v:"finalizers.fluxcd.io": {}
    manager: helm-controller
    operation: Update
    time: "2026-03-03T15:30:35Z"
  - apiVersion: helm.toolkit.fluxcd.io/v2
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        f:conditions: {}
        f:failures: {}
        f:helmChart: {}
        f:history: {}
[...]

PR #1430, opened 2026-03-04
Fix enqueing the same revision while reconciling

matheuscscp:
Fixes: #1429

We traced the behavior reported in the issue back to the enqueue functions: they skip enqueue when the object is Ready=True and the revision matches. This should also happen if Reconciling=True. This check works because the object is patched with the attempted revision on the beginning of the reconciliation. Note that there's still a (very) small window where the object is not patched yet in which the issue can still happen.

### DIFF
--- a/internal/controller/helmrelease_indexers.go
+++ b/internal/controller/helmrelease_indexers.go
@@ -30,6 +30,10 @@ import (
 	v2 "github.com/fluxcd/helm-controller/api/v2"
 )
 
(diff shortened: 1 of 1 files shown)


matheuscscp:
@hiddeco No worries, there's a bug in upstream Helm that is staged for 4.1.2 next Wednesday. A Flux patch release is scheduled for the day after. I already did some testing and will do more in any case, but if by the release you have time for another peek it would be most welcome :pray:

PR #1447, opened 2026-03-19
Support helm release upgrade on helm chart name change

### LINKED ISSUE #870: Can't change chart name without triggering uninstall (flux 2.2.x)
We were relying on ability to change chart name (`spec.chart.spec.chart` field) in existing HelmRelease without deletion of underlying resources. The charts in question are similar internally, but are stored with different names. This worked fine in flux 2.1.x and older, but since 2.2.x we noticed that changing chart name would always trigger uninstall and re-install from scratch, which causes problems.

Controller log contains the following line:

```
release target configuration changed (chart name): running uninstall for current release
```

It looks like these are responsible code parts:
https://github.com/fluxcd/helm-controller/blob/v0.37.2/internal/controller/helmrelease_controller.go#L321
https://github.com/fluxcd/helm-controller/blob/v0.37.2/internal/action/verify.go#L69

It would be great if there was at least an option to disable uninstall in such scenario, especially since it was working perfectly fine before. It would be helpful even in simple scenarios where chart maintainer decided to change chart name, and sudden uninstall can lead to nasty side effects, especially if chart installs CRDs. I'm not in the clear why such dependency is even needed, given that release name is not tied to chart name.

MichaelMorrisEst:
### Summary
Closes #870 

Proposes to add a new property in the 'upgrade' element of the HelmRelease CRD.

The new property shall have two allowed values: "InPlaceUpdate" and "Reinstall"

When the value "Reinstall" is specified, or when no value is specified, the current behaviour of uninstalling a Helm Release when the Helm chart name is changed shall remain unchanged. This results in a new Helm Release install.
When the value "InPlaceUpdate" is specified, the Helm Release will not be uninstalled when the Helm chart name is changed. This results in a Helm upgrade being performed. This was the behaviour before Flux 2.2.0

This re-introduces support that was lost going from Flux 2.1.x -> 2.2.x but controls the behaviour through the HelmRelease custom resource so either the previous (<2.2.x) behaviour or current behaviour can be achieved

### Changes

- Add `chartNameChangeStrategy` property to the 'Upgrade' struct in helmrelease_types.go and ensure the default "Reinstall" is used if no value is specified
- In helmrelease_controller.go if the chart name has changed, only proceed to uninstall the HelmRelease if the new property is set to "Reinstall"
- Change the visibility of `action.targetChartName` to enable it to be referenced in the HelmRelease controller
- Tests added to verify new behaviour

### Original proposal
The original proposal when the PR was submitted was to introduce a feature gate to control the behaviour. Following discussions in the comments below the implementation changed to use the custom resource instead. The above description has been updated to reflect the new implementation

### DIFF
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -790,6 +790,16 @@ const (
 	ServerSideApplyAuto ServerSideApplyMode = "auto"
 )
 
+// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes
+type ChartNameChangeStrategy string
+
+const (
+	// Update the Helm release in place.
+	InPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
+	// Reinstall the Helm release, uninstalling the existing Helm release
+	Reinstall ChartNameChangeStrategy = "Reinstall"
+)
+
 // Upgrade holds the configuration for Helm upgrade actions for this
 // HelmRelease.
 type Upgrade struct {
@@ -886,6 +896,16 @@ type Upgrade struct {
 	// +kubebuilder:validation:Enum=enabled;disabled;auto
 	// +optional
 	ServerSideApply ServerSideApplyMode `json:"serverSideApply,omitempty"`
+
+	// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes.
+	// Valid values are 'Reinstall' or 'InPlaceUpdate'. Defaults to 'Reinstall' if omitted.
+	//
+	// Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
+	//
+	// InPlaceUpdate: Update the Helm release in place.
+	//
+	// +optional
+	ChartNameChangeStrategy ChartNameChangeStrategy `json:"chartNameChangeStrategy,omitempty"`
 }
 
 // GetTimeout returns the configured timeout for the Helm upgrade action, or the
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -795,9 +795,9 @@ type ChartNameChangeStrategy string
 
 const (
 	// Update the Helm release in place.
-	InPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
+	ChartNameChangeStrategyInPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
 	// Reinstall the Helm release, uninstalling the existing Helm release
-	Reinstall ChartNameChangeStrategy = "Reinstall"
+	ChartNameChangeStrategyReinstall ChartNameChangeStrategy = "Reinstall"
 )
 
 // Upgrade holds the configuration for Helm upgrade actions for this
@@ -904,6 +904,7 @@ type Upgrade struct {
 	//
 	// InPlaceUpdate: Update the Helm release in place.
 	//
+	// +kubebuilder:validation:Enum=InPlaceUpdate;Reinstall
 	// +optional
 	ChartNameChangeStrategy ChartNameChangeStrategy `json:"chartNameChangeStrategy,omitempty"`
 }
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -918,6 +918,13 @@ func (in Upgrade) GetTimeout(defaultTimeout metav1.Duration) metav1.Duration {
 	return *in.Timeout
 }
 
+func (in Upgrade) GetChartNameChangeStrategy() ChartNameChangeStrategy {
+	if in.ChartNameChangeStrategy == "" {
+		return ChartNameChangeStrategyReinstall
+	}
+	return in.ChartNameChangeStrategy
+}
+
 // GetRemediation returns the configured Remediation for the Helm upgrade
 // action.
 func (in Upgrade) GetRemediation() Remediation {
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -794,9 +794,9 @@ const (
 type ChartNameChangeStrategy string
 
 const (
-	// Update the Helm release in place.
+	// ChartNameChangeStrategyInPlaceUpdate updates the Helm release in place.
 	ChartNameChangeStrategyInPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
-	// Reinstall the Helm release, uninstalling the existing Helm release
+	// ChartNameChangeStrategyReinstall reinstalls the Helm release, uninstalling the existing Helm release.
 	ChartNameChangeStrategyReinstall ChartNameChangeStrategy = "Reinstall"
 )
 
--- a/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
+++ b/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
@@ -919,6 +919,15 @@ spec:
                 description: Upgrade holds the configuration for Helm upgrade actions
                   for this HelmRelease.
                 properties:
+                  chartNameChangeStrategy:
+                    description: |-
+                      ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes.
+                      Valid values are 'Reinstall' or 'InPlaceUpdate'. Defaults to 'Reinstall' if omitted.
+
+                      Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
+
+                      InPlaceUpdate: Update the Helm release in place.
+                    type: string
                   cleanupOnFail:
                     description: |-
                       CleanupOnFail allows deletion of new resources created during the Helm
--- a/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
+++ b/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
@@ -927,6 +927,9 @@ spec:
                       Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
 
                       InPlaceUpdate: Update the Helm release in place.
+                    enum:
+                    - InPlaceUpdate
+                    - Reinstall
                     type: string
                   cleanupOnFail:
                     description: |-
--- a/internal/action/verify.go
+++ b/internal/action/verify.go
@@ -43,10 +43,10 @@ var (
 )
 
 const (
-	targetStorageNamespace = "storage namespace"
-	targetReleaseNamespace = "release namespace"
-	targetReleaseName      = "release name"
-	TargetChartName        = "chart name"
+	TargetStorageNamespace = "storageNamespace"
+	TargetReleaseNamespace = "releaseNamespace"
+	TargetReleaseName      = "releaseName"
+	TargetChartName        = "chartName"
 )
 
 // ReleaseTargetChanged returns a reason and true if the given release and/or
@@ -62,11 +62,11 @@ func ReleaseTargetChanged(obj *v2.HelmRelease, chartName string) (string, bool) 
 	case obj.Status.StorageNamespace == "", cur == nil:
 		return "", false
 	case obj.GetStorageNamespace() != obj.Status.StorageNamespace:
-		return targetStorageNamespace, true
+		return TargetStorageNamespace, true
 	case obj.GetReleaseNamespace() != cur.Namespace:
-		return targetReleaseNamespace, true
+		return TargetReleaseNamespace, true
 	case release.ShortenName(obj.GetReleaseName()) != cur.Name:
-		return targetReleaseName, true
+		return TargetReleaseName, true
 	case chartName != cur.ChartName:
 		return TargetChartName, true
 	default:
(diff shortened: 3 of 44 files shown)


internal/features/features.go
```diff
@@ -132,6 +137,9 @@ var features = map[string]bool{
 	// DefaultToRetryOnFailure
 	// opt-in from v1.5.2
 	DefaultToRetryOnFailure: false,
+	// UninstallOnChartNameChange
+	// opt-in from v1.5.4
```
hiddeco, line 141: The chance of this going into a patch release is unlikely, this should thus indicate 1.6.0.

internal/action/verify.go
```diff
@@ -46,7 +46,7 @@ const (
 	targetStorageNamespace = "storage namespace"
 	targetReleaseNamespace = "release namespace"
 	targetReleaseName      = "release name"
-	targetChartName        = "chart name"
+	TargetChartName        = "chart name"
```
hiddeco, line 49: This list of constants was always meant to be a descriptive reason (factored out to a const for code readability), not an identifier. Additionally, only exposing _one_ of them creates a weak contract. I would prefer a solution where:

1. All identifiers are public
2. Identifiers do not contain spaces

hiddeco (commented):
Feedback beyond the contents of the PR: I do wonder if there are not other changes (e.g., release name/namespace or storage namespace) where there could be a desire of control. Given this, I do wonder if what this PR attempts to do isn't too narrow.

hiddeco:
Some additional thoughts:

- For storage namespace changes, we could actually do the more intelligent thing and attempt to _migrate_ instead of going through an uninstall cycle. This would open a surface for "Secret smuggling" however, and integrity would likely have to be verified as part of this.
- In hindsight, the biggest design mistake I ever introduced in the helm-controller (and something that still bites me as an operator once in a while) is not having the `storageNamespace` field mirror the `releaseNamespace` unless explicitly overwritten. If we are rethinking how this all works, this might be worth addressing as well (behind a feature gate).
- Some of these changes may actually _require_ an uninstall (or juggling with Helm install options). As changing a release name would make the resources stay put, which could then cause the subsequent Helm install to fail.

artem-nefedov:
How about having this setting on resource level rather than a global feature flag?
I feel like that's more correct approach, but don't know about the complexity involved to support that.

MichaelMorrisEst:
> How about having this setting on resource level rather than a global feature flag? I feel like that's more correct approach, but don't know about the complexity involved to support that.

Hi @artem-nefedov Are you suggesting some on the CRD? e.g. maybe adding something to the 'Uninstall' element of 'HelmRelease'?

MichaelMorrisEst:
Hi @hiddeco 
Thats for your feedback.

I dont think Helm itself supports changing the helm release name, namespace or storage namespace and, as you state yourself, this might require an uninstall anyway. The proposed scope of chart name changes only I believe is consistent with what Helm supports. I'd be concerned about the complexity of implementing this for other changes particularly if there isnt anybody looking for it. Having said that, I am happy to work on it if you feel its the way to go, but I would need your advice on how it should be implemented.

In relation to the default `storageNamespace`. If I understand you correctly, you are saying it would be better for it to default to `targetNamespace` and this change should be controlled by a feature gate? I am happy to implement such a change, but would it be better in a separate PR as its not directly related to this proposal?

artem-nefedov:
> Hi @artem-nefedov Are you suggesting some on the CRD? e.g. maybe adding something to the 'Uninstall' element of 'HelmRelease'?

Sorry for the delayed reply. Yes, I meant something like this, only rather than 'Uninstall', I think the better place is 'Upgrade'. For example:

```yaml
spec:
  upgrade:
    chartNameChangePolicy: InPlaceUpdate # or "Reinstall" (default)
```

MichaelMorrisEst:
Hi @hiddeco 
If you get a chance to look at my replay above and let me know how you would like to proceed that would be great, thanks

hiddeco:
I would be okay with:

```yaml
spec:
  upgrade:
    onChartNameChange: InPlaceUpdate # or "Reinstall" (default)
```

(or `Upgrade` and `Reinstall`).

matheuscscp:
> I would be okay with:
> 
> ```yaml
> spec:
>   upgrade:
>     onChartNameChange: InPlaceUpdate # or "Reinstall" (default)
> ```

I think we have a tendency to use `somethingStrategy`

hiddeco:
That would also work for me. In any case, I am fine with allowing this to be defined at HelmRelease object level.

config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
```diff
@@ -918,6 +918,15 @@ spec:
...
                 properties:
+                  chartNameChangeStrategy:
+                    description: |-
+                      ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes.
+                      Valid values are 'Reinstall' or 'InPlaceUpdate'. Defaults to 'Reinstall' if omitted.
+
+                      Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
+
+                      InPlaceUpdate: Update the Helm release in place.
+                    type: string
```
artem-nefedov, line 933: should have enum with allowed values

api/v2/helmrelease_types.go
```diff
@@ -886,6 +896,16 @@ type Upgrade struct {
...
 	// +optional
 	ServerSideApply ServerSideApplyMode `json:"serverSideApply,omitempty"`
+
+	// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes.
+	// Valid values are 'Reinstall' or 'InPlaceUpdate'. Defaults to 'Reinstall' if omitted.
+	//
+	// Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
+	//
+	// InPlaceUpdate: Update the Helm release in place.
+	//
```
artem-nefedov, line 906: should have kubebuilder option for enum validation - see `ServerSideApply` as example

api/v2/helmrelease_types.go
```diff
@@ -790,6 +790,16 @@ const (
...
 )
 
+// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes
+type ChartNameChangeStrategy string
+
+const (
+	// Update the Helm release in place.
+	InPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
+	// Reinstall the Helm release, uninstalling the existing Helm release
+	Reinstall ChartNameChangeStrategy = "Reinstall"
```
artem-nefedov: consider prefixing names with a type

internal/action/verify.go
MichaelMorrisEst: Done

config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
MichaelMorrisEst: Done

api/v2/helmrelease_types.go
MichaelMorrisEst: Done

api/v2/helmrelease_types.go
MichaelMorrisEst: Done

MichaelMorrisEst:
Hi @artem-nefedov @hiddeco 
Just checking if you have any further comments on this PR or if you happy with the updates

internal/controller/helmrelease_controller.go
```diff
@@ -372,7 +372,7 @@ func (r *HelmReleaseReconciler) reconcileRelease(ctx context.Context,
 	// If the release target configuration has changed, we need to uninstall the
 	// previous release target first. If we did not do this, the installation would
 	// fail due to resources already existing.
-	if reason, changed := action.ReleaseTargetChanged(obj, loadedChart.Name()); changed {
+	if reason, changed := action.ReleaseTargetChanged(obj, loadedChart.Name()); changed && (reason != action.TargetChartName || obj.Spec.Upgrade.ChartNameChangeStrategy == v2.ChartNameChangeStrategyReinstall) {
```
artem-nefedov: This condition explicitly checks for "Reinstall", but I don't understand where default value is supposed to be set as "Reinstall".
Is there a test to confirm the default behavior?

MichaelMorrisEst: My bad, the setting of the default value got lost when I switched from feature gate to resource level setting. Added back in now and test to validate

internal/action/verify.go
```diff
@@ -43,10 +43,10 @@ var (
...
 
 const (
-	targetStorageNamespace = "storage namespace"
-	targetReleaseNamespace = "release namespace"
-	targetReleaseName      = "release name"
-	targetChartName        = "chart name"
+	TargetStorageNamespace = "storageNamespace"
+	TargetReleaseNamespace = "releaseNamespace"
+	TargetReleaseName      = "releaseName"
+	TargetChartName        = "chartName"
```
artem-nefedov: It looks like those are also used as a part of human-readable log message, so I don't know if it is correct to change the format.
The maintainers can decide.
I don't see any more obvious problems with the PR.

MichaelMorrisEst:
Hi @hiddeco, @matheuscscp 
If you can spare some time to review that would be great

api/v2/helmrelease_types.go
```diff
@@ -790,6 +790,16 @@ const (
...
 )
 
+// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes
+type ChartNameChangeStrategy string
+
+const (
+	// Update the Helm release in place.
+	ChartNameChangeStrategyInPlaceUpdate ChartNameChangeStrategy = "InPlaceUpdate"
+	// Reinstall the Helm release, uninstalling the existing Helm release
+	ChartNameChangeStrategyReinstall ChartNameChangeStrategy = "Reinstall"
```
matheuscscp: The Go standard for comments on symbols is starting the comment with the symbol name and add full stop (`.`) at the end.

internal/action/verify.go
matheuscscp: Well pointed, I think we can just make the symbols public so the calling code can use them, but since they are used in a log I think we need to keep the literals as they are.

matheuscscp (commented):
@MichaelMorrisEst Please add a few tests here: https://github.com/fluxcd/helm-controller/blob/main/.github/workflows/e2e.yaml

api/v2/helmrelease_types.go
MichaelMorrisEst: Done

internal/action/verify.go
MichaelMorrisEst: Done

MichaelMorrisEst:
> @MichaelMorrisEst Please add a few tests here: https://github.com/fluxcd/helm-controller/blob/main/.github/workflows/e2e.yaml

Hi @matheuscscp In order to add tests there I would need a renamed version of podinfo chart somewhere. Do you think it would be ok to add something to the pofinfo repo (or create a new repo?), or do you see some other approach?

matheuscscp:
> > @MichaelMorrisEst Please add a few tests here: https://github.com/fluxcd/helm-controller/blob/main/.github/workflows/e2e.yaml
> 
> Hi @matheuscscp In order to add tests there I would need a renamed version of podinfo chart somewhere. Do you think it would be ok to add something to the pofinfo repo (or create a new repo?), or do you see some other approach?

@MichaelMorrisEst You can do it like the CRD tests at the end, they run from a chart managed in this repo.

MichaelMorrisEst:
Hi @matheuscscp do you have any further comments?

matheuscscp (approved):
LGTM! 🚀

MichaelMorrisEst:
Thanks @matheuscscp for taking the time to review. Do you think it will be possible to get this merged in time to be included in the upcoming Flux 2.9 release?

matheuscscp:
Depends on another maintainer stamping the PR

MichaelMorrisEst:
Hi @stefanprodan @hiddeco if you could spare some time to review it would be much appreciated. It would be great to get this in Flux 2.9 if possible.

api/v2/helmrelease_types.go
```diff
@@ -886,6 +896,17 @@ type Upgrade struct {
...
+	// ChartNameChangeStrategy defines the strategy to use when a Helm chart name changes.
+	// Valid values are 'Reinstall' or 'InPlaceUpdate'. Defaults to 'Reinstall' if omitted.
+	//
+	// Reinstall: Reinstall the Helm release, uninstalling the existing Helm release.
+	//
+	// InPlaceUpdate: Update the Helm release in place.
+	//
+	// +kubebuilder:validation:Enum=InPlaceUpdate;Reinstall
+	// +optional
+	ChartNameChangeStrategy ChartNameChangeStrategy `json:"chartNameChangeStrategy,omitempty"`
```
stefanprodan, line 909: How would users know about this? Please write docs in: https://github.com/fluxcd/helm-controller/blob/main/docs/spec/v2/helmreleases.md

stefanprodan (changes requested):

api/v2/helmrelease_types.go
MichaelMorrisEst: Documentation added

stefanprodan:
Please rebase with upstream main branch and force push.

stefanprodan (approved):
LGTM

Thanks @MichaelMorrisEst 🏅

matheuscscp (approved):
LGTM! 🚀

PR #1451, opened 2026-03-30
Route chart name changes through upgrade instead of uninstall/reinstall

mattnewell:
## Summary

- Chart name changes in a HelmRelease currently trigger a full uninstall/reinstall cycle via `ReleaseTargetChanged()`, causing unnecessary downtime by deleting all resources before recreating them
- Unlike release name or namespace changes, a chart name change does not affect the Helm storage key — `helm upgrade` correctly diffs old and new manifests, removing resources no longer in the new chart
- `VerifyRelease()` already detects chart name changes via `ErrChartChanged`, routing them to `ReleaseStatusOutOfSync` and the upgrade path — so the fallback behavior is already in place

## Test plan

- [x] Updated `TestReleaseTargetChanged` to assert chart name change returns `false`
- [x] Full test suite passes (`make test`)
- [ ] Integration test with a live cluster: change chart name on an existing HelmRelease and verify the controller performs an upgrade (not uninstall/reinstall) and existing resources are preserved

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/action/verify.go
+++ b/internal/action/verify.go
@@ -67,8 +67,6 @@ func ReleaseTargetChanged(obj *v2.HelmRelease, chartName string) (string, bool) 
 		return targetReleaseNamespace, true
 	case release.ShortenName(obj.GetReleaseName()) != cur.Name:
 		return targetReleaseName, true
-	case chartName != cur.ChartName:
-		return targetChartName, true
 	default:
 		return "", false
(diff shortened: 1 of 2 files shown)


hiddeco:
Duplicate of #1447. Please refer to the discussion there.

PR #1455, opened 2026-04-09
Ignore force replace when server-side apply is enabled

### LINKED ISSUE #1453: Update with force: true fails
(Presumably since Helm 2.8.0 or later), upgrading a HelmRelease with 

````
upgrade:
  force: true
````

fails with this error: `invalid operation: cannot use force conflicts and force replace together`. This seems to come out of helm: https://github.com/helm/helm/blob/main/pkg/kube/client.go#L818C33-L818C105

A possible train of thought (just grepping through the source code): 
- Since some time (1.5.0), helm-controller defaults to server-side apply. (from Changelog).
- That sets forceConflicts always to true. (https://github.com/fluxcd/helm-controller/blob/8a0f9d044fbec659f88db3e60a028901ab42b7ec/internal/action/upgrade.go#L77)
- "upgrade.force: true" sets forceReplace also to true (https://github.com/fluxcd/helm-controller/blob/8a0f9d044fbec659f88db3e60a028901ab42b7ec/internal/action/upgrade.go#L119)
-  and that triggers that check in Helm.

matheuscscp:
Fixes: #1453

### DIFF
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -810,7 +810,10 @@ type Upgrade struct {
 	// +optional
 	DisableSchemaValidation bool `json:"disableSchemaValidation,omitempty"`
 
-	// Force forces resource updates through a replacement strategy.
+	// Force forces resource updates through a replacement strategy
+	// that avoids 3-way merge conflicts on client-side apply.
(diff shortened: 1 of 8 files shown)


stefanprodan (approved):
LGTM

PR #1460, opened 2026-04-17
Fix post-renderer conflict of overlapping hooks and templates

### LINKED ISSUE #1449: Helm release with pre-install hook causes upgrades to break with "post render on files may not add resource with an already registered id"
### Describe the bug

Since upgrading to 2.8 we are seeing cases of Helm upgrades failing

`Helm upgrade failed for release default/app with chart app@1.0.0: error while running post render on files: accumulating resources: accumulation err=’accumulating resources from ‘helm-output.yaml’: may not add resource with an already registered id: ServiceAccount.v1.[noGrp]/app.[noNs]‘: must build at directory: ‘/helm-output.yaml’: file is not directory`

The service account created by hook is a one-time install time resource, as seen in annotations:
```
    "helm.sh/hook": pre-install
    "helm.sh/hook-weight": "-10"
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
So hook job can run with SA on the first install. Release SA takes over after that for upgrades.

Seems the post render doesn't take that into account.

### Steps to reproduce

Have a Helm chart create service account annotated as above. On upgrade observe failure. I'm not sure exactly how to trigger failure, because its not happening for all charts yet.

### Expected behavior

Upgrades completing without post-render failure from kustomize assuming that all resources will be created at the same time

### Screenshots and recordings

_No response_

### OS / Distro

EKS

### Flux version

2.8.3

### Flux check

N/A

### Git provider

_No response_

### Container Registry provider

_No response_

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

matheuscscp:
Fixes: #1449
Includes: https://github.com/helm/helm/pull/32049

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,11 +4,8 @@ go 1.26.0
 
 replace github.com/fluxcd/helm-controller/api => ./api
 
-// TODO: Remove this after Helm 4.1.4 comes out.
-// This fixes the following bugs:
-//   - https://github.com/helm/helm/issues/31867
-//   - https://github.com/helm/helm/issues/31935
-replace helm.sh/helm/v4 => github.com/fluxcd/helm/v4 v4.1.4-flux.1
+// TODO: Remove this after Helm 4.2 comes out.
+replace helm.sh/helm/v4 => github.com/fluxcd/helm/v4 v4.1.4-flux.3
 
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM 🚀

PR #1470, opened 2026-04-24
Add post-render strategy support and conditional defaults

### LINKED ISSUE #1469: Feature Request: Support all the post-render strategies available in the Helm SDK
**⚠️ Breaking change**: From Flux 2.8.0 to 2.8.5 the default was `combined` (Helm 4 breaking change), then in Flux 2.8.6 we fixed the regression (changed the behavior to `nohooks`). In Flux 2.9 (via #1470) we are changing this default back to `combined` for honoring the feature gate `UseHelm3Defaults`, which defaults to the **Helm 4** defaults when the gate is `false`, and to the **Helm 3** defaults when the gate is `true`. We are sustaining our decision to keep aligned with Helm's defaults throughout Flux's and Helm's lives. This commitment implies shipping breaking changes whenever Helm changes default functionality in major versions.

**The behavior can also be pinned to a specific value directly in the `HelmRelease` object through the field introduced in #1470.**

Helm 4.2 will introduce 3 post-render strategies (see https://github.com/helm/helm/issues/31985 and https://github.com/helm/helm/pull/32049):

* `nohooks`: Helm 3 behavior, hooks are not sent to post-renderers. (this is Flux's behavior in >=2.8.6 <2.9.0, which was broken in >=2.8.0 <2.8.6)
* `combined`: Helm 4's default behavior, both hooks and templates are sent to post-renderers in the same stream. (this was the broken behavior in Flux >=2.8.0 <2.8.6)
* `separate`: Opt-in from Helm 4.2, hooks and templates are sent to post-renderers in separate streams.

I propose we introduce `HelmRelease.spec.postRenderStrategy` as an enum matching the 3 values defined by Helm, then in the install and upgrade actions this value can be passed to the SDK. The default should be `nohooks`, which is the current behavior in Flux >=2.8.6 <2.9.0.

vocarista:
Fixes #1469

**⚠️ Breaking change**: From Flux 2.8.0 to 2.8.5 the default was `combined` (Helm 4 breaking change), then in Flux 2.8.6 we fixed the regression (changed the behavior to `nohooks`). In Flux 2.9 (through this PR) we are changing this default back to `combined` for honoring the feature gate `UseHelm3Defaults`, which defaults to the **Helm 4** defaults when the gate is `false`, and to the **Helm 3** defaults when the gate is `true`. We are sustaining our decision to keep aligned with Helm's defaults throughout Flux's and Helm's lives. This commitment implies shipping breaking changes whenever Helm changes default functionality in major versions.

**The behavior can also be pinned to a specific value directly in the `HelmRelease` object through the field introduced in this PR.**

Adds support for Helm's post-render strategies (nohooks, combined, separate) being introduced in Helm 4.2. The new `spec.postRenderStrategy` field allows users to configure how hooks are sent to post-renderers.

**Changes:**
- Add `PostRenderStrategy` enum with values: `nohooks`, `combined`, `separate`
- Add `spec.postRenderStrategy` field to `HelmReleaseSpec`
- Add `toHelmPostRenderStrategy()` mapping function in action package
- Update install and upgrade actions to use the configured strategy
- Add unit tests for default behaviour and all strategy values

**Default behavior:**
- Defaults to `combined` to align with upstream Helm
- Respects `UseHelm3Defaults` feature gate by switching to `nohooks` when enabled

### DIFF
--- a/internal/action/post_render.go
+++ b/internal/action/post_render.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action
+
+import (
+	"helm.sh/helm/v4/pkg/action"
+
+	v2 "github.com/fluxcd/helm-controller/api/v2"
+)
+
+// toHelmPostRenderStrategy converts the API PostRenderStrategy to the Helm SDK value.
+// If the strategy is not set, it defaults to PostRenderStrategyCombined (Helm 4 default),
+// or PostRenderStrategyNoHooks when UseHelm3Defaults is enabled.
+func toHelmPostRenderStrategy(strategy v2.PostRenderStrategy) action.PostRenderStrategy {
+	if strategy == "" {
+		if UseHelm3Defaults {
+			return action.PostRenderStrategyNoHooks
+		}
+		return action.PostRenderStrategyCombined
+	}
+	return action.PostRenderStrategy(strategy)
+}
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -185,6 +185,15 @@ type HelmReleaseSpec struct {
 	// +optional
 	PostRenderers []PostRenderer `json:"postRenderers,omitempty"`
 
+	// PostRenderStrategy defines the strategy for sending hooks to post-renderers.
+	// Valid values are 'nohooks' (hooks not sent to post-renderers, Helm 3 behavior),
+	// 'combined' (hooks and templates sent together, Helm 4 default), and 'separate'
+	// (hooks and templates sent in separate streams, Helm 4.2 opt-in).
+	// Defaults to 'combined', or 'nohooks' when the UseHelm3Defaults feature gate is enabled.
+	// +kubebuilder:validation:Enum=nohooks;combined;separate
+	// +optional
+	PostRenderStrategy PostRenderStrategy `json:"postRenderStrategy,omitempty"`
+
 	// WaitStrategy defines Helm's wait strategy for waiting for applied
 	// resources to become ready.
 	// +optional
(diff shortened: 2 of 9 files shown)


internal/action/ssa.go
```diff
@@ -33,3 +35,23 @@ func toHelmSSAValue(mode v2.ServerSideApplyMode) string {
 		return string(mode)
 	}
 }
+
+// toHelmPostRenderStrategy converts the API PostRenderStrategy to the Helm SDK value.
+// If the strategy is not set, it defaults to PostRenderStrategyCombined (Helm 4 default),
+// or PostRenderStrategyNoHooks when UseHelm3Defaults is enabled.
+func toHelmPostRenderStrategy(strategy v2.PostRenderStrategy) action.PostRenderStrategy {
```
matheuscscp: I think we can create a new file for this. All the `UseHelm3Defaults`-aware features should have their own dedicated file (the other two already have)

matheuscscp (approved):
Just a nit, otherwise LGTM! 🚀

internal/action/ssa.go
matheuscscp: Make it `post_render.go`

matheuscscp: Please amend and force-push :pray:

vocarista: Did it, had to do another push because I'd initially named it postrender_strategy.go before I saw your latest comment.

matheuscscp:
@vocarista Sorry, I forgot something very important, please add the field docs to `docs/spec/v2/helmreleases.md` :pray:

matheuscscp:
Then please squash the commits into one and force-push :pray:

vocarista:
Squashed 👾

matheuscscp:
@vocarista Can you please open a PR here too? https://github.com/fluxcd/website/blob/main/content/en/flux/components/helm/options.md?plain=1#L67

Then a second PR here (highlighting the breaking change):
https://github.com/fluxcd/website/blob/main/content/en/roadmap.md?plain=1#L41

They have to be two separate PRs because the second one will go live immediately

vocarista:
@matheuscscp I've opened the PRs on fluxcd/website

internal/action/post_render.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	switch strategy {
+	case v2.PostRenderStrategyCombined:
+		return action.PostRenderStrategyCombined
+	case v2.PostRenderStrategySeparate:
+		return action.PostRenderStrategySeparate
+	default:
+		return action.PostRenderStrategyNoHooks
+	}
+}
```
hiddeco, line 36: Nit.

But given we do validate at CRD level, I wonder if this simply can't be:


```suggestion
func toHelmPostRenderStrategy(strategy v2.PostRenderStrategy) action.PostRenderStrategy {
    if strategy == "" {
        if UseHelm3Defaults {
            return action.PostRenderStrategyNoHooks
        }
        return action.PostRenderStrategyCombined
    }
    return action.PostRenderStrategy(strategy)
}
```

or

```suggestion
func toHelmPostRenderStrategy(strategy v2.PostRenderStrategy) action.PostRenderStrategy {
    switch strategy {
    case "":
        if UseHelm3Defaults {
            return action.PostRenderStrategyNoHooks
        }
        return action.PostRenderStrategyCombined
    case v2.PostRenderStrategyCombined:
        return action.PostRenderStrategyCombined
    case v2.PostRenderStrategySeparate:
        return action.PostRenderStrategySeparate
    default:
        return action.PostRenderStrategyNoHooks
    }
}
```

vocarista: Thanks @hiddeco, applied!

matheuscscp (approved):
LGTM! 🚀 

@vocarista Thanks very much!

PR #1473, opened 2026-04-25
docs: add debugging tips to DEVELOPMENT.md

alliasgher:
## Summary
Adds a 'Debugging the controller locally' section to `DEVELOPMENT.md` covering knobs not already in 'How to run the controller locally':

- `RUNTIME_NAMESPACE` to scope the watch
- `--concurrent=1` to serialize reconciles for a clean trace
- `flux suspend` for unrelated objects sharing the cluster

Cross-controller setup is already documented (or n/a for this controller) and is not duplicated here.

Assisted-by: Claude/claude-opus-4-7

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -98,3 +98,41 @@ Deploy `helm-controller` into the cluster that is configured in the local kubeco
 ```sh
 make deploy
 ```
+
+## Debugging the controller locally
+
+When reproducing an issue or stepping through reconciliation logic, the
+following knobs make local runs cheaper and the resulting logs easier to
+read.
+
+### Limit the watched namespace
+
+The controller watches every namespace by default. To narrow it to a single
+namespace, set the `RUNTIME_NAMESPACE` environment variable before invoking
+`make run`:
+
+```sh
+RUNTIME_NAMESPACE=flux-system make run
+```
+
+### Reduce reconcile concurrency
+
+Each `HelmRelease` reconcile is processed concurrently (default
+`--concurrent=4`). When debugging it is almost always easier to follow a
+serial trace; pass `--concurrent=1` so reconciles run one at a time:
+
+```sh
+go run ./main.go --concurrent=1
+```
+
+### Suspend unrelated objects
+
+If the controller is sharing a cluster with other Flux objects, suspend
+anything not relevant to the test you're running so their reconciles don't
+interleave with yours:
+
+```sh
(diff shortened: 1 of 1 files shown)


alliasgher:
Self-correcting this before it wastes a reviewer's time: two of the tips in this PR are wrong, and I only found out because @stefanprodan reviewed the same text on the kustomize-controller version (fluxcd/kustomize-controller#1646).

**The `RUNTIME_NAMESPACE` tip does not work.** That variable is only read when `--watch-all-namespaces` is false:

```go
if !watchOptions.AllNamespaces {
    watchNamespace = os.Getenv("RUNTIME_NAMESPACE")
}
```

and the flag defaults to `true` (`fluxcd/pkg/runtime/controller/watch.go`), so `RUNTIME_NAMESPACE=flux-system make run` on its own changes nothing. I verified this controller has the identical code. Stefan also made the broader point that narrowing the watch would hide cross-namespace behaviour you often need while debugging, so the section should go rather than be corrected.

**The `--concurrent=1` tip is backwards** as general debugging advice: goroutine leaks, deadlocks and races only surface under concurrency, and it is redundant anyway once unrelated objects are suspended.

That is most of the substance, so I do not think there is a useful PR left here. Happy to close it, or to cut it down to just the accurate remainder if a maintainer thinks that is worth keeping. Sorry for the noise, and for sending the same wrong text to several repos at once.

alliasgher:
Closing this. Two of the three tips were wrong rather than merely imprecise, and once they are removed there is not enough left to justify a review cycle.

For the record, so this is searchable if anyone tries the same thing later:

- **Limit the watched namespace** was wrong. `RUNTIME_NAMESPACE` is only read when `--watch-all-namespaces` is false, and that flag defaults to `true`, so the documented command had no effect. @stefanprodan also made the stronger point that narrowing the watch would hide cross-namespace behaviour you often need while debugging, so the section should not exist even in a corrected form.
- **Reduce reconcile concurrency** was backwards as general advice: goroutine leaks, deadlocks and races only surface under concurrency, and it is redundant once unrelated objects are suspended.
- **Suspend unrelated objects** was the only sound tip, and it amounts to mentioning that `flux suspend <kind> --all` exists, which is not worth a PR on its own.

Thanks @stefanprodan for the review that caught this, and apologies to the other repos for receiving the same wrong text. If a maintainer does want a debugging section in DEVELOPMENT.md, I am happy to write one from scratch against what the flags actually do.

PR #1475, opened 2026-04-27
Add optional SOPS decryption for HelmRelease values

SebTardif:
## End goal
This PR contributes to closing [fluxcd/flux2#4075](https://github.com/fluxcd/flux2/issues/4075): **"HelmRelease should support SOPS encryption directly"**.

The target user outcome is that SOPS-encrypted HelmRelease values can be used without forcing users to split YAMLs or move secrets into separate formats just to satisfy schema validation/workarounds.

## What this PR does (scope)
This PR implements the **helm-controller side** of that goal:

- adds `spec.decryption` to HelmRelease API
- decrypts composed Helm values in controller reconcile flow when decryption is configured
- adds unit coverage for decryption logic
- includes integration-test scaffolding and regenerated API/CRD/docs artifacts

In short: this PR is the runtime/controller capability needed for HelmRelease value decryption.

## Relation to other repos / dependency clarification
There is related work in flux CLI: [fluxcd/flux2#5869](https://github.com/fluxcd/flux2/pull/5869), including maintainer feedback in [this comment](https://github.com/fluxcd/flux2/pull/5869#issuecomment-4329018725) about an opt-in `--strip-sops-metadata` flag.

That flux2 change improves CLI UX (notably `flux diff`) for manifests containing SOPS metadata. This PR is complementary but addresses a different layer.

- **Not a hard runtime dependency:** helm-controller decryption support can be reviewed/merged independently.
- **Complementary UX dependency:** to fully address issue #4075 end-to-end from both controller behavior and CLI ergonomics, both PRs are useful together.

## Current PR status
Keeping this PR as **draft** while review feedback is incorporated.


stefanprodan:
In Flux, kustomize-controller is responsible for decrypting the HelmRelease values. What’s in this PR can not happen as it breaks Flux security model.  

Such a major change requires an RFC sponsored by a core maintainer.

PR #760, opened 2023-08-24
Support other Helm storage backends besides Secrets

fiscafusca:
## Overview
This PR adds support for all Helm storage backends (see the [official documentation](https://helm.sh/docs/topics/advanced/#storage-backends)).

## Usage
The storage driver can be set through the `--helm-storage-driver` flag, and its value is propagated to the `NewRunner` function. The set of allowed values is [`secret`, `configmap`, `sql`]. If the flag is unset, the value is retrieved from the environment variable `HELM_DRIVER`. If the environment variable is also unset, the value simply defaults to `secret` - for backwards compatibility.

## Use case
The possibility to switch between backends is essential to have more flexibility in the cluster. Storing release information in Secrets may not always be the best option, mostly due to:
- Size of Helm releases information, which may exceed 1MB;
- Combined total size of cluster Secrets, which may cause disruptions in the cluster;
- Helm release location - in some use cases it could be necessary to move release information to another location for compliance.

Moving the release information, for example, to a SQL backend, would easily address these issues and allow keeping a longer history of deployments.

## Testing (cluster)
The changes were tested on a local K8s cluster. I personally used a single node `kind` cluster with K8s `v1.27.3`.

### Steps to reproduce
Create the cluster:
```bash
kind create cluster
```

Build a local image (I named it `test-helm-controller:latest`) and load it to the kind cluster:
```bash
export IMG=test-helm-controller:latest
make docker-build
kind load image test-helm-controller:latest
```

Deploy the default `helm-controller` and `source-controller`:
```bash
make deploy
```

Create the namespace for the Helm release:
```bash
kubectl create namespace hello
```

Prepare the manifests for the `HelmRepo` and `HelmRelease` to use for testing. I personally used [this chart](https://artifacthub.io/packages/helm/cloudecho/hello) for simplicity.

`helmrepo.yaml`:
```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: HelmRepository
metadata:
  name: cloudecho
  namespace: hello
spec:
  interval: 1m0s
  url: https://cloudecho.github.io/charts/
```
`helmrelease.yaml`:
```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: hello
  namespace: hello
spec:
  chart:
    spec:
      chart: hello
      sourceRef:
        kind: HelmRepository
        name: cloudecho
      version: 0.1.2
  values:
    replicaCount: 3
  interval: 1m0s
```

[...]

PR #1479, opened 2026-05-05
Support configurable Helm storage driver (Secret / ConfigMap / Memory / SQL)

### LINKED ISSUE #272: [Feature Request] Support SQL storage backend
Helm supports [SQL storage backend](https://helm.sh/docs/topics/advanced/#sql-storage-backend), it would be nice to support it in helm-controller, it would be useful especially when not able to create secret or configmap on kubernetes cluster. Here's a real situation we meet https://github.com/fluxcd/flux2/discussions/1473

funkypenguin:
## Overview

Adds a `--helm-storage-driver` flag and `HELM_DRIVER` environment-variable fallback that select the Helm release storage backend, mirroring the Helm CLI's `HELM_DRIVER`. Supported values: `secret`/`secrets`, `configmap`/`configmaps`, `memory`, `sql` — matched case-insensitively. Unset preserves current behaviour (Secret), so the change is fully backwards compatible.

The SQL driver reads its connection string from `HELM_DRIVER_SQL_CONNECTION_STRING` (matching Helm CLI).

This PR is a fresh take on the long-stalled #760 (open since 2023). I rebased it against current `main`, ported it to the post-Helm-v4 / `ConfigFactory` era, and addressed several issues uncovered during a careful review:

- **Pool lifecycle**: a single `helmdriver.NewSQL` per reconcile would leak a connection pool indefinitely (Helm v4 has no `Close` on `storage.Driver`). The new `action.SQLDriverPool` caches drivers per storage namespace, bounding live pools to the set of namespaces actually in use rather than once per reconcile.
- **Startup probe**: when `--helm-storage-driver=sql`, the controller probes the database with a transient `database/sql.Open` + `PingContext` + `Close` before starting the manager, surfacing invalid DSNs or unreachable backends at boot rather than failing every HelmRelease reconcile. The probe connection is cleanly closed and does not contribute to the long-lived pool.
- **DSN sanitisation**: `helmdriver.NewSQL` / `sqlx.Connect` errors can echo the DSN. They're caught at the `WithStorage` call site and replaced with a generic message so DSN material cannot leak into `HelmRelease` status conditions.
- **Helm CLI parity**: `secret`/`secrets` and `configmap`/`configmaps` are accepted interchangeably, matching `pkg/action.(*Configuration).Init`.
- **DI pattern**: SQL test fakes go through a per-pool `SetFactory` rather than a package-level mutable global, matching the existing `ConfigFactoryOption` idiom in this package.

## Use cases

- Helm release information size exceeds the 1 MiB Secret limit (large values, large CRD-heavy charts).
- Cluster-wide pressure from cumulative Secret count.
- Compliance requirements to store release data outside the cluster.

## Backwards compatibility

- No flag/env changes: continues to use the Secret driver. `HELM_DRIVER` is now consulted as a fallback (it was previously ignored).
- The `--helm-storage-driver` flag value is normalised once at startup; an invalid value fails the controller process, not every reconcile.

## Tests

Adds coverage for:
- Driver name normalisation (case-insensitive, alias handling).
- SQL driver pool: per-namespace caching, factory error propagation.
- Sanitised SQL connection errors (assertion that the DSN is not surfaced).
- The "no SQL pool configured" error path.
- `IsSupportedStorageDriver` / `NormalizeStorageDriverName`.

`go test ./internal/action -run 'TestWithStorage|TestSQLDriverPool|TestIsSupportedStorageDriver|TestConfigFactory|TestNewConfigFactory|TestWithDriver|TestWithSQLDriverPool|TestStorageLog'` passes (all 30+ subtests, no envtest required).

## References

- Closes #272 (long-standing request for non-Secret backends).
- Supersedes #760 (same feature, ported to current code, with the lifecycle/UX/security gaps that came up in that PR's discussion addressed).

CC @fiscafusca (original #760 author) — thank you for the groundwork; the configuration UX in this PR follows yours.

🤖 The implementation was assisted by Claude Code; the resulting code, tests, and design decisions are my own and have been reviewed by an independent Codex pass before submission.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/google/cel-go v0.26.1
 	github.com/google/go-cmp v0.7.0
 	github.com/hashicorp/go-retryablehttp v0.7.8
+	github.com/lib/pq v1.10.9
 	github.com/mitchellh/copystructure v1.2.0
 	github.com/onsi/gomega v1.39.1
 	github.com/opencontainers/go-digest v1.0.1-0.20231025023718-d50d2fec9c98
@@ -157,7 +158,6 @@ require (
 	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/lann/builder v0.0.0-20180802200727-47ae307949d0 // indirect
 	github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0 // indirect
-	github.com/lib/pq v1.10.9 // indirect
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mailru/easyjson v0.9.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
--- a/internal/action/config.go
+++ b/internal/action/config.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strings"
+	"sync"
 
 	helmaction "helm.sh/helm/v4/pkg/action"
 	helmstorage "helm.sh/helm/v4/pkg/storage"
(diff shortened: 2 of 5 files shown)


matheuscscp:
Nice work! But I think we need a bit of discussion for this feature, e.g. here you are supporting it only at the controller level. I can see value in creating spec fields instead. Would you be able to join the open flux dev meetings to discuss this? https://fluxcd.io/community/

PR #1485, opened 2026-05-15
Move retryingRoundTripper wrapping to constructor

### LINKED ISSUE #1484: ToRESTConfig stacks retryingRoundTripper on every call
## Bug description

`MemoryRESTClientGetter.ToRESTConfig()` calls `c.cfg.Wrap()` unconditionally on every invocation. Because `rest.Config.Wrap()` is additive and the config is shared across the getter's lifetime, each call stacks another `retryingRoundTripper` layer around the transport.

The retrying round-tripper was added in #1084 to fix temporary `etcdserver: leader changed` errors from kube-apiserver. However, the `Wrap` call was placed inside `ToRESTConfig()` without a guard, so it fires on every call rather than once.

## Impact

- Each reconciliation invokes `ToRESTConfig()` multiple times (via `ToDiscoveryClient`, `ToRESTMapper`, and direct calls).
- Over time, the nesting depth grows linearly: after N calls, each HTTP request traverses N retry layers.
- This causes memory growth from retained closure/wrapper chains and deepening call stacks.

## Steps to reproduce

1. Create a `MemoryRESTClientGetter` with a `*rest.Config`.
2. Call `ToRESTConfig()` 10 times.
3. Inspect `cfg.WrapTransport(nil)` and walk the `retryingRoundTripper` chain.
4. Observe 10 nested layers instead of 1.

## Expected behavior

The transport should be wrapped exactly once regardless of how many times `ToRESTConfig()` is called.

## Proposed fix

Guard the `c.cfg.Wrap()` call with a `sync.Once` field on the struct, consistent with the existing lazy-init patterns already used in the struct (`sync.Mutex` for discoveryClient, restMapper, clientCfg).

SebTardif:
## Description

`MemoryRESTClientGetter.ToRESTConfig()` calls `c.cfg.Wrap()` unconditionally
on every invocation. Because `rest.Config.Wrap()` is additive, each call
stacks another `retryingRoundTripper` layer around the transport. Over time
this causes unbounded nesting: after N calls, each HTTP request traverses N
retry layers, growing memory and deepening call stacks.

The retrying round-tripper was added in #1084 to fix temporary
`etcdserver: leader changed` errors from kube-apiserver. The wrapping logic
was correct, but placing it inside `ToRESTConfig()` without a guard means it
fires on every call rather than once.

Each reconciliation invokes `ToRESTConfig()` multiple times (via
`ToDiscoveryClient`, `ToRESTMapper`, and direct calls). Over time the
nesting depth grows linearly, causing:

- Memory growth from retained closure/wrapper chains
- Deepening call stacks on every HTTP round-trip
- Unnecessary overhead per request

## Changes

- Add a `sync.Once` field (`wrapOnce`) to `MemoryRESTClientGetter` to
  ensure the transport is wrapped exactly once, consistent with the
  existing lazy-init patterns (`sync.Mutex`) used for `discoveryClient`,
  `restMapper`, and `clientCfg` in the same struct.
- Add a test that calls `ToRESTConfig()` 10 times and asserts exactly
  one `retryingRoundTripper` layer exists in the `WrapTransport` chain.

Fixes #1484

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -112,6 +112,11 @@ func NewMemoryRESTClientGetter(cfg *rest.Config, opts ...Option) *MemoryRESTClie
 		opts(g)
 	}
 	g.setDefaults()
+	// Add retries to fix temporary "etcdserver: leader changed" errors from kube-apiserver.
+	// Done here once rather than in ToRESTConfig to avoid stacking wrappers on repeated calls.
+	g.cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return &retryingRoundTripper{wrapped: rt}
+	})
 	return g
 }
 
@@ -131,10 +136,6 @@ func (c *MemoryRESTClientGetter) ToRESTConfig() (*rest.Config, error) {
 	if c.cfg == nil {
 		return nil, fmt.Errorf("MemoryRESTClientGetter has no REST config")
 	}
-	// add retries to fix temporary "etcdserver: leader changed" errors from kube-apiserver
-	c.cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
-		return &retryingRoundTripper{wrapped: rt}
-	})
 	return c.cfg, nil
 }
 
--- a/internal/kube/client_test.go
+++ b/internal/kube/client_test.go
@@ -184,6 +184,30 @@ func TestMemoryRESTClientGetter_ToRESTConfig(t *testing.T) {
 		g.Expect(cfg).To(BeIdenticalTo(c.cfg))
 	})
 
+	t.Run("wraps transport only once across multiple calls", func(t *testing.T) {
+		g := NewWithT(t)
+
+		c := NewMemoryRESTClientGetter(&rest.Config{
+			Host: "https://example.com",
+		})
+
+		// Call ToRESTConfig multiple times.
+		for i := 0; i < 10; i++ {
+			_, err := c.ToRESTConfig()
+			g.Expect(err).ToNot(HaveOccurred())
+		}
+
+		// Verify that only one retryingRoundTripper layer exists by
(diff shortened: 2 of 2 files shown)


internal/kube/client.go
```diff
@@ -132,8 +134,10 @@ func (c *MemoryRESTClientGetter) ToRESTConfig() (*rest.Config, error) {
 		return nil, fmt.Errorf("MemoryRESTClientGetter has no REST config")
 	}
 	// add retries to fix temporary "etcdserver: leader changed" errors from kube-apiserver
-	c.cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
-		return &retryingRoundTripper{wrapped: rt}
+	c.wrapOnce.Do(func() {
+		c.cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+			return &retryingRoundTripper{wrapped: rt}
+		})
```
matheuscscp: Why can't this just be done once in the constructor without any concurrency primitives? It doesn't look like `c.cfg` is being updated anywhere during the lifetime of `*MemoryRESTClientGetter`, so there's no reason for doing this change in `c.cfg` lazily at all, if I understand this code correctly

SebTardif:
Thanks for the review @matheuscscp. You're right, there's no reason for this to be lazy. `cfg` is fully configured after the options loop in the constructor and is never modified afterward. I moved the `Wrap` call to the end of `NewMemoryRESTClientGetter` and removed `sync.Once` entirely.

matheuscscp:
Thanks @SebTardif, I updated the PR title. Please squash into one commit, make the commit msg match the PR title and force-push :pray:

SebTardif:
Done. Squashed into a single commit matching the PR title.

matheuscscp (approved):
LGTM! 🚀 

Thanks @SebTardif!

PR #1491, opened 2026-05-16
Add missing defer summarize(req) in UninstallRemediation

### LINKED ISSUE #1488: UninstallRemediation missing defer summarize(req), Ready condition not updated
## Description

`UninstallRemediation.Reconcile` is the only action reconciler that does not call `defer summarize(req)`. All other reconcilers include this deferred call:

- `Install` (install.go:79)
- `Upgrade` (upgrade.go:75)
- `Test` (test.go:78)
- `Uninstall` (uninstall.go:88)
- `Unlock` (unlock.go:64)
- `RollbackRemediation` (rollback_remediation.go:82)

The `summarize` function propagates sub-conditions (`Released`, `Remediated`, `TestSuccess`) into the top-level `Ready` condition and refreshes `ObservedPostRenderersDigest` / `ObservedCommonMetadataDigest`.

`UninstallRemediation` is a `ReconcilerTypeRemediate`, so after it runs, `AtomicRelease.strategy.MustStop()` returns `true` and the outer loop exits without its own summarize call. This means the `Ready` condition is never updated after an uninstall remediation, leaving stale status until the next reconciliation pass.

## Steps to Reproduce

1. Configure a HelmRelease with uninstall remediation strategy
2. Trigger a failed upgrade that activates uninstall remediation
3. Observe that the `Ready` condition is not updated after remediation completes

## Expected Behavior

The `Ready` condition should be updated after uninstall remediation, consistent with all other reconcilers.

SebTardif:
## Description

`UninstallRemediation.Reconcile` is the only action reconciler that does not call `defer summarize(req)`. All six other reconcilers (`Install`, `Upgrade`, `Test`, `Uninstall`, `Unlock`, `RollbackRemediation`) include this call.

Without it, the `Ready` condition is never updated after an uninstall remediation, and `ObservedPostRenderersDigest` / `ObservedCommonMetadataDigest` are not refreshed.

Closes #1488

## Changes

One-line addition: `defer summarize(req)` after the variable block, matching all other reconcilers.

## Verification

- `go build ./...` passes
- `go vet ./...` passes
- Integration tests require etcd (CI)

### DIFF
--- a/internal/reconcile/uninstall_remediation.go
+++ b/internal/reconcile/uninstall_remediation.go
@@ -84,6 +84,8 @@ func (r *UninstallRemediation) Reconcile(ctx context.Context, req *Request) erro
 		cfg    = r.configFactory.Build(logBuf, observeUninstall(req.Object, v2.ReleaseActionUninstallRemediation))
 	)
 
+	defer summarize(req)
+
 	// Require current to run uninstall.
 	if cur == nil {
 		return fmt.Errorf("%w: required to uninstall", ErrNoLatest)
--- a/internal/reconcile/uninstall_remediation_test.go
+++ b/internal/reconcile/uninstall_remediation_test.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/fluxcd/pkg/apis/meta"
 	"github.com/fluxcd/pkg/runtime/conditions"
 
 	"github.com/fluxcd/pkg/chartutil"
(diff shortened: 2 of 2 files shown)


hiddeco:
Thanks for the patch. 

Double checked things, and `rollback_remediation.go` and `uninstall_remediation.go` were added in the same commit (b975b3f), with the same doc comment, but only the former got `defer summarize(req)`. This makes me believe it was indeed missed (and is a bug).

However, before this can be merged, please add a test in `uninstall_remediation_test.go` that asserts `Ready`'s reason equals `UninstallSucceeded` after a successful run to prevent regressions in the future.

Also: if this work was assisted by an LLM, please add the appropriate trailer to the commit per https://github.com/fluxcd/helm-controller?tab=contributing-ov-file#ai-coding-assistants-guidance.

SebTardif:
Updated with ReadyCondition assertions on all test paths (success and failure), matching the pattern in rollback_remediation_test.go. The Assisted-by trailer was already present.

Issue #1486, opened 2026-05-16
Bug: crds: CreateReplace fails when chart's crds/ directory contains non-CRD resources (e.g. ValidatingAdmissionPolicy)


p-zany:
Upstream: https://github.com/envoyproxy/gateway/issues/9015

## Bug description

When a Helm chart bundles non-CRD resources (such as `ValidatingAdmissionPolicy`) inside its `crds/` directory (or a sub-chart's `crds/` directory), setting `crds: CreateReplace` on a `HelmRelease` causes the install/upgrade to fail with an error like:

```
failed to apply CustomResourceDefinitions: failed to update CustomResourceDefinition(s): no ValidatingAdmissionPolicy with the name "safe-upgrades.gateway.networking.k8s.io" found
```

## Reproduction

Using `gateway-helm` chart from `oci://docker.io/envoyproxy/gateway-helm` at version `1.8.0`. This chart bundles `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` resources inside `charts/crds/crds/gatewayapi-crds.yaml` (the `crds/` directory of the `crds` sub-chart).

HelmRelease:
```yaml
spec:
  install:
    crds: CreateReplace
  upgrade:
    crds: CreateReplace
```

Error on install when the `ValidatingAdmissionPolicy` does not yet exist in the cluster:
```
failed to apply CustomResourceDefinitions: failed to update CustomResourceDefinition(s):
no ValidatingAdmissionPolicy with the name "safe-upgrades.gateway.networking.k8s.io" found
```

## Root cause

`internal/action/crds.go` in the `CreateReplace` branch processes all resources found in a chart's `crds/` directories. When it encounters a non-`CustomResourceDefinition` resource (like `ValidatingAdmissionPolicy`), it attempts an `update` (PUT) operation on it. If the resource does not yet exist in the cluster, the update fails with "not found" instead of falling back to a create.

## Expected behavior

`CreateReplace` should handle non-CRD resources found in `crds/` directories using upsert semantics (create if not exists, replace if exists) — the same as Server-Side Apply — rather than failing when the resource is absent.

Alternatively, non-CRD resources in `crds/` could be skipped with a warning, since placing them there is unconventional (see the related upstream issue in envoyproxy/gateway).

## Environment

- flux2: v2.x (helm-controller using `helm.toolkit.fluxcd.io/v2`)
[...]

PR #1494, opened 2026-05-16
Ignore non-CRD objects under crds/

matheuscscp:
xref: https://github.com/fluxcd/helm-controller/issues/1486#issuecomment-4466918319


stefanprodan (approved):
LGTM

PR #1515, opened 2026-06-12
Update fluxcd/pkg dependencies

fluxcdbot:
Automated `fluxcd/pkg` dependency update for `main`.

Includes Helm 4.2.1.

Generated by: https://github.com/fluxcd/helm-controller/actions/runs/27430483789


matheuscscp (approved):

matheuscscp:
Note: Helm 4.2.1 is changing a behavior in their kstatus implementation that affects the test I fixed in the commit above. This change affects only the test, not helm-controller! helm-controller uses flux's own kstatus implementation (shared with kustomize-controller).

PR #1526, opened 2026-06-23
Add categories to the HelmRelease CRD

josesiqueira:
This adds the standard Flux resource categories to the HelmRelease CRD in this
repository so users can list Flux resources with kubectl, for example:

```
kubectl get fluxcd          # all Flux resources
kubectl get fluxcd-appliers   # all resources in this category
kubectl get all             # now includes Flux resources
```

Each CRD gets exactly three categories:

```
all
fluxcd
fluxcd-appliers
```

The `+kubebuilder:resource` markers were removed from the skipped older
API versions, since they interfere with CRD generation, and the CRDs
were regenerated with `make manifests`.

Part of fluxcd/flux2#5947

### DIFF
--- a/api/v2/helmrelease_types.go
+++ b/api/v2/helmrelease_types.go
@@ -1431,7 +1431,7 @@ const (
 
 // +genclient
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:shortName=hr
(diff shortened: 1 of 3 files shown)


matheuscscp (approved):
LGTM! 🚀

Issue #1546, opened 2026-07-22
Helm-controller gets state wrong when before-hook-creation fails with 403


dee0sap:
We have chart X which was deployed successfully.
Then a developer added a test hook.   Snippet from helm status below gives you an idea of what it looks like
```yaml
 - delete_policies:
      - before-hook-creation
      - hook-succeeded
    events:
      - test
    kind: ClusterRoleBinding
    last_run:
      phase: ""
```
For sake of discussion say this was the release number where this was rolled out was release 5.

The ServiceAccount used to deploy chart X had privileges for create and listing ClusterRoleBindings and privileges for deleting certain named ClusterRoleBindings but not the new test hook ClusterRoleBinding.

Release 5 had a status of 'deployed' since the actual upgrade wasn't a problem.

When the helm-controller tried to run the tests however the delete for 'before-hook-creation' failed.

In this case helm's pkg/action/release_testing.go executes this snippet 
```go
	shutdown, err := r.cfg.execHookWithDelayedShutdown(rel, release.HookTest, kube.StatusWatcherStrategy, r.WaitOptions, r.Timeout, serverSideApply)

	if err != nil {
		rel.Hooks = append(skippedHooks, rel.Hooks...)
		r.cfg.Releases.Update(reli)
		return reli, shutdown, err
	}
```

The above `Update` call will end up triggering the callback below which is in the helm-controller's internal/reconcile/test.go
```go
func observeTest(obj *v2.HelmRelease) storage.ObserveFunc {
	return func(rlsr helmrelease.Releaser) {
		rls, ok := rlsr.(*helmreleasev1.Release)
		if !ok {
			return
		}
		// Only accept test results for the latest release.
		if !obj.Status.History.Latest().Targets(rls.Name, rls.Namespace, rls.Version) {
			return
		}

		// Update the latest snapshot with the test result.
		latest := obj.Status.History.Latest()
		tested := release.ObservedToSnapshot(releaseToObservation(rls, latest, latest.GetAction()))
		tested.SetTestHooks(release.TestHooksFromRelease(rls))
		obj.Status.History[0] = tested
	}
}
```

At this point the snapshot will look something like what is below.   *Note* the `testHook` entries with no `LastStarted`, `LastCompleted` or `Phase` value.
```yaml
  history:
    - action: upgrade
      apiVersion: v2
      appVersion: v2.8.0
      chartName: sok-eso
      chartVersion: 1.1.135-52.db0b227git+c393abec5dfd
[...]

PR #1547, opened 2026-07-22
Fix `HasBeenTested` for all corner cases

dee0sap:
xref: #1546

Updated test status handling so a release only counts as tested when all selected test hooks have reached a terminal phase. Empty selected hook sets still count as tested for charts with no matching tests. **Filtered-out hooks are no longer recorded in the snapshot**, avoiding false incomplete results from intentionally skipped tests.

### DIFF
--- a/api/v2/snapshot_types.go
+++ b/api/v2/snapshot_types.go
@@ -49,6 +49,9 @@ const (
 	// snapshotTestPhaseFailed indicates that the test of the release the snapshot
 	// was taken from has failed.
 	snapshotTestPhaseFailed = "Failed"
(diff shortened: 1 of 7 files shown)


matheuscscp:
If I understand correctly, this fix implies that a non-empty list of zero TestHooks still means the release was not tested, but it's also semantically different from an empty list of TestHooks? Why is that?

stefanprodan (approved):
LGTM

dee0sap:
Hey @matheuscscp 

Thanks for the review and the question.

I was wanting to just address the problem my team was having without having to worry about any subtleties of how HasBeenTested is used.

Also thanks @stefanprodan :)

PR #1554, opened 2026-07-27
Drop controller-runtime dependency from API module

alvaroaleman:
This change removes the usage of the deprecated controller-runtime scheme builder from the api module, thereby getting rid of the entire controller-runtime dependency in there.


matheuscscp (approved):
LGTM! 🚀

PR #1558, opened 2026-08-02
fix: preserve YAML anchors after Helm v4 List unwrap in post-render

### LINKED ISSUE #1556: YAML anchors/aliases spanning resources break during post-render: "unknown anchor referenced"
### Describe the bug

A chart template that defines a YAML anchor (`&name`) on one resource and references it with an alias (`*name`) on another resource in the same rendered output fails to reconcile on helm-controller **v1.6.x**. Reconciliation errors out during the **post-render** stage with:

```
error while running post render on files: MalformedYAMLError: yaml: unknown anchor 'jobSpec' referenced
```

The same chart:
- renders correctly with `helm template` (anchors are preserved as text and resolve when the manifest is parsed as a single stream), and
- installed/upgraded correctly on older helm-controller **v1.0.x** (Helm SDK v3).

It looks like the post-render pipeline parses/re-serializes the manifest per-document (or otherwise splits the anchor definition from the alias reference), so the alias can no longer find its anchor.

### Steps to reproduce

1. A chart with a template that shares a spec between two resources via a YAML anchor. Minimal, abstract example:

   ```yaml
   # templates/example.yaml
   apiVersion: v1
   kind: List
   items:
     - apiVersion: batch/v1
       kind: Job
       metadata:
         name: example
       spec: &jobSpec
         template:
           spec:
             restartPolicy: Never
             containers:
               - name: main
                 image: busybox:latest
                 command: ["true"]
     - apiVersion: batch/v1
       kind: CronJob
       metadata:
         name: example
       spec:
         schedule: "0 0 * * *"
         jobTemplate:
           spec: *jobSpec        # <-- alias reference
   ```

2. A HelmRelease that installs this chart, with **no** `postRenderers`, `commonMetadata`, etc. configured:

   ```yaml
   kind: HelmRelease
   metadata:
     name: example
     namespace: default
   spec:
     interval: 5m
     chart:
       spec:
         chart: example
         version: 0.1.0
         sourceRef:
           kind: HelmRepository
           name: example
   ```

3. Reconcile.

### Expected behavior

The alias resolves to the anchored node (as it does with `helm template` and as it did on helm-controller v1.0.x), and the release installs successfully.

### Actual behavior

Reconciliation fails during post-render:

```
Helm install failed for release default/example with chart example@0.1.0:
error while running post render on files: MalformedYAMLError: yaml: unknown anchor 'jobSpec' referenced
```

Note: the post-render stage runs even though the HelmRelease configures **no** post-renderer.

### Additional context

- The failure is in the **post-render** stage, before apply. Setting `spec.install.serverSideApply: false` and `spec.upgrade.serverSideApply: disabled` (client-side apply) makes **no difference** — the release never reaches the apply stage. So this is unrelated to the client-side/server-side apply change.
- Verified by deleting and re-creating the HelmRelease from scratch (new object, fresh generation): the very first reconcile fails identically, so it is not stale/cached state.
- Reproduced on **both v1.6.2 and v1.6.3** (confirmed by running `ghcr.io/fluxcd/helm-controller:v1.6.3` directly — identical failure), so it is not fixed in the latest release.
- This is a behavior change between helm-controller v1.0.x (Helm SDK v3) and v1.6.x (Helm SDK v4). It may originate in the upstream Helm SDK's post-render handling rather than helm-controller itself — filing here since that is where the regression surfaced; happy to move it upstream if that is the right place.

A named template (`define`/`include`) is a working post-render-safe replacement for the anchor, so there is a workaround. Filing because YAML anchors are valid, previously worked, and the failure mode ("unknown anchor") is confusing to diagnose.

### Steps to reproduce — versions

- helm-controller version: **v1.6.2** (also present on latest **v1.6.3**)
- flux2 (community chart): 2.19.0 (appVersion 2.9.1)
- Kubernetes version: v1.34.x (k3s)

zanarellidev:
## Summary
- Helm v4 `annotateAndMerge` unwraps `kind: List` into separate documents while keeping `&anchor` / `*alias` markers, which turns valid in-List aliases into cross-document references.
- Origin labels always run as a post-renderer, so `NewResMapFromBytes` fails with `MalformedYAMLError: yaml: unknown anchor ... referenced` even when the HelmRelease has no `postRenderers`.
- Inflate aliases in `Combined.Run` only when parse fails with an unknown-anchor error: re-wrap documents as a List, `DeAnchor`, re-serialize.

Fixes #1556

### Bisect
- Reproduced on current `main`.
- OriginLabels alone has failed on raw cross-doc aliases since at least `v1.0.0`.
- The List-unwrap path that produces those aliases for a chart like the one in #1556 arrives with **Helm v4** (`v1.5.0`: Helm `v4.1.x`; `v1.6.0`: Helm `v4.2.x`). Helm v3 controllers did not unwrap List this way before post-render.

## Test plan
- [x] `go test -count=1 ./internal/postrender/`
- [x] Fail-before / pass-after on `Test_OriginLabels_Run_crossDocAnchorsFromListUnwrap` (exact `unknown anchor 'jobSpec'` without inflate)

### DIFF
--- a/internal/postrender/anchors.go
+++ b/internal/postrender/anchors.go
@@ -1,0 +1,125 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package postrender
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/kio"
+)
+
+// inflateYAMLAliases expands YAML anchors/aliases in a multi-document stream.
+//
+// Helm v4's post-render merge (annotateAndMerge) unwraps kind:List into
+// individual documents while preserving YAML anchors. That turns in-List
+// aliases into cross-document references, which standard YAML parsers reject
+// ("unknown anchor referenced"). Re-wrapping into a List restores a single
+// document scope so aliases resolve, then DeAnchor inlines them permanently.
+//
+// If the input already parses cleanly, it is returned unchanged to avoid
+// reformatting manifests.
+func inflateYAMLAliases(in []byte) ([]byte, error) {
+	if len(bytes.TrimSpace(in)) == 0 {
+		return in, nil
+	}
+
+	if _, err := kio.ParseAll(string(in)); err == nil {
+		return in, nil
+	} else if !isUnknownAnchorErr(err) {
+		return nil, err
+	}
+
+	wrapped := wrapDocsAsList(string(in))
+	nodes, err := kio.ParseAll(wrapped)
+	if err != nil {
+		return nil, fmt.Errorf("inflate YAML aliases: %w", err)
+	}
+	for _, n := range nodes {
+		if err := n.DeAnchor(); err != nil {
+			return nil, fmt.Errorf("inflate YAML aliases: %w", err)
+		}
+	}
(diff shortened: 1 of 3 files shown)


matheuscscp:
Looks like the bug is in helm, not helm-controller

PR #1565, opened 2026-08-26
Move v1.6 back to upstream helm

matheuscscp:
Helm 4.2.4 includes https://github.com/helm/helm/pull/32327, so we can move back.

The main branch never moved to the fork.


stefanprodan (approved):
🚀

PR #1574, opened 2026-08-31
Ignore NotFound when deleting HelmChart

### LINKED ISSUE #1566: reconcileDelete should ignore HelmChart Delete NotFound after cached Get
Hi maintainers,

I think there may be a narrow delete-result handling issue in the HelmChart template cleanup path.

In `reconcileDelete`, the controller fetches the referenced `HelmChart`. If the get returns `NotFound`, cleanup continues because the chart is already absent:

[`internal/reconcile/helmchart_template.go:191-L198`](https://github.com/fluxcd/helm-controller/blob/6dfc0849abf22b6da826f07936bd701263b32a40/internal/reconcile/helmchart_template.go#L191-L198)

If the get succeeds, the controller deletes the chart. This delete branch returns any delete error, including `NotFound`:

[`internal/reconcile/helmchart_template.go:199-L204`](https://github.com/fluxcd/helm-controller/blob/6dfc0849abf22b6da826f07936bd701263b32a40/internal/reconcile/helmchart_template.go#L199-L204)

A possible interleaving is:

```text
the controller reads HelmChart H
H is already deleted by a previous reconcile attempt or another controller action
the cache has not observed that deletion yet
reconcileDelete still enters the delete branch for H
Delete(ctx, H) reaches the API server and returns NotFound
reconcileDelete returns an error instead of clearing obj.Status.HelmChart
```

For this cleanup path, absence of the referenced HelmChart seems to be the desired state. Would it be reasonable to ignore `apierrors.IsNotFound(err)` from the delete call after the cached get has found a chart, while still returning other delete failures?

Chao

ChaoHuang2018:
Fixes #1566.

`reconcileDelete` already treats `NotFound` from the initial HelmChart `Get` as completed cleanup. This change applies the same behavior to the following `Delete` call.

If the HelmChart is removed after `Get` succeeds but before `Delete` reaches the API server, the cleanup has still completed and `Status.HelmChart` can be cleared.

Changes:

- ignore `NotFound` from `r.client.Delete(ctx, &chart)`
- add a regression test for this race

### DIFF
--- a/internal/reconcile/helmchart_template.go
+++ b/internal/reconcile/helmchart_template.go
@@ -198,7 +198,7 @@ func (r *HelmChartTemplate) reconcileDelete(ctx context.Context, obj *v2.HelmRel
 		}
 		if err == nil {
 			// Delete the HelmChart.
-			if err = r.client.Delete(ctx, &chart); err != nil {
+			if err = r.client.Delete(ctx, &chart); client.IgnoreNotFound(err) != nil {
 				err = fmt.Errorf("failed to delete HelmChart '%s': %w", obj.Status.HelmChart, err)
 				return err
 			}
--- a/internal/reconcile/helmchart_template_test.go
+++ b/internal/reconcile/helmchart_template_test.go
@@ -28,7 +28,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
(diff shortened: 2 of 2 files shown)


matheuscscp:
Hi @ChaoHuang2018, this looks good, please rebase and force-push so I can merge :pray:

ChaoHuang2018:
Thanks! I rebased onto the latest main and force-pushed.

matheuscscp (approved):
LGTM

Issue #1588, opened 2026-09-16
SSA turns the default CRD policy into an upsert, CRDs get applied on every upgrade


dmichel1:
since v1.5.0, HelmReleases default to server side apply and chart CRDs now get updated on
every upgrade. before that they were left alone once they existed. this isn't in the
v1.5.0 changelog and the docs still describe the old behavior.

Couple of questions:

1. are the CRDs being force applied on upgrade intended, or a side effect of the SSA default?
1. if intended, update the docs? `.spec.upgrade.crds` is documented as defaulting to `Skip`:

   https://github.com/fluxcd/helm-controller/blob/f37ac2aecf07f83e70bb9179e7c6c5053f7ddfeb/docs/spec/v2/helmreleases.md#L656

## what's happening

`applyCRDs` hasn't changed much, but the `Create` branch now passes the SSA option through:

https://github.com/fluxcd/helm-controller/blob/372c34e2b4a455d24ef6cbb8321c82c4533dbbd8/internal/action/crds.go#L147-L150

client side that's a Create, hits AlreadyExists, logs "is already present. Skipping."
server side it's an apply, so it updates whatever is already there. `ForceConflicts` is
always on for SSA:

https://github.com/fluxcd/helm-controller/blob/372c34e2b4a455d24ef6cbb8321c82c4533dbbd8/internal/action/upgrade.go#L77

so there's no conflict surfaced, it just takes ownership and writes.

## repro

kind cluster, ack eks controller chart, 1.16.3 -> 1.23.0, no `crds` field set anywhere.

| helm-controller | apply method | CRDs updated |
|---|---|---|
| v1.4.3 (flux 2.7.3) | client side | no |
| v1.6.3 (flux 2.9.4) | SSA (default) | yes |
| v1.6.3 (flux 2.9.4) | `serverSideApply: disabled` | no |

only difference between the last two rows is SSA. on the default run the CRD generation
went 1 -> 2 and managedFields picked up `manager=helm-controller op=Apply`. 4 of the 10
CRDs changed, matching the diff between the two chart versions.

matheuscscp:
@dmichel1 Thanks for this issue, it just uncovered a churny history behind  `.spec.upgrade.crds`.

#738 changed the default from Skip to Create, probably by mistake since there are no mentions about the change in the PR, so it's a regression.

But I don't think we want to change the default again. It's been a long time since #738, so at this point we should just fix the docs to match reality.

[...]

PR #1589, opened 2026-09-17
Fix updating CRDs on upgrades with the Create policy when using SSA

matheuscscp:
CSA yields a conflict error in this scenario, SSA does not. This PR Get's the CRD first to emulate the same error, and lets the following "if already exists error" code branch kick in like it does for CSA.

xref: #1588 (Avoiding `Fixes` to keep issue open until a patch release is out)


stefanprodan (approved):
LGTM

